PR #11, opened 2019-07-22
Automated tests and CI for Ingest Pipeline

eweitz:
This adds automated tests for extraction and transformation code in Ingest Pipeline.  They verify that inputs to our load-to-Firestore code look as expected.  This helps protect against regressions in a way that requires much less manual work.

Mocks are used for test speed and isolation.  Code coverage -- i.e. tested percent of lines and branches -- is also measured.   These tests can also be parallelized, and tests can be run selectively based on string matching on the test name.

See "EXAMPLES" in `tests/test_ingest.py` for more on basic usage and those special test features.

I have also set up continuous integration to run these tests in CircleCI:

https://circleci.com/gh/broadinstitute/scp-ingest-pipeline

This satisfies [SCP-1730](https://broadworkbench.atlassian.net/browse/SCP-1730).


tests/test_ingest.py
```diff
@@ -0,0 +1,154 @@
...
+    this method with expected argument values.
+
+    Ideally we would mock Firestore itself, but existing libraries to mock
+    Firestore (e.g. https://github.com/mdowds/python-mock-firestore) don't
+    implement important Firestore methods like "batch()".
+    """
+    self.load_expression_data_args = args
+    self.load_expression_data_kwargs = kwargs
+
+def mock_firestore_client(*args, **kwargs):
```
knapii-developments: I don't think this function is required. You're not expecting a returned value, if anything you're testing to see that it was called with the correct arguments.  I think you could do something like firestore.Client = Mock() and mock the whole object. That way when you call any method of firestore.Client it would just returned a mocked object.

tests/test_ingest.py
```diff
@@ -0,0 +1,154 @@
...
+    return actual_model, expected_model
+
+# Mock method that loads data to Firestore
+IngestPipeline.load_expression_data = mock_load_expression_data
+
+
+class IngestTestCase(unittest.TestCase):
+
+    def setup_ingest(self, args):
+        args_list = args.split(' ')
```
knapii-developments: Here is where firestore.Client() = Mock would live

ingest/ingest_pipeline.py
```diff
@@ -191,22 +191,29 @@ def parse_arguments():
...
-    parser_ingest_cluster = subargs.add_parser('ingest_clusters',
+    parser_ingest_cluster = subparser.add_parser('ingest_clusters',
                                                help='Indicates that cluster'
                                                ' files are being ingested')
     parser_ingest_cluster.add_argument('--cluster-file',
                                        help='Path to cluster files')
-    parsed_args = args.parse_args()
+    return parser
+
+def validate_arguments(parsed_args):
```
knapii-developments: Here you're mock the args argument. Then you'd set the return values of parsed_args.matrix_file_type.return_value to an incorrect value.  Then you'd check to see if the assertion error was raised with assertRaises.

knapii-developments: You'd also test for the positive test case here as well.

tests/test_ingest.py
```diff
@@ -0,0 +1,154 @@
...
+        arguments = vars(parsed_args)
+
+        ingest = IngestPipeline(**arguments)
+
+        if hasattr(ingest, 'ingest_expression'):
+            getattr(ingest, 'ingest_expression')()
+
+        return ingest
+
+    @patch('google.cloud.firestore.Client', side_effect=mock_firestore_client)
```
knapii-developments: So here you would have 3 cases (give or take a test or 2) for dense matrices. One for ingesting dense matrices correctly (test_ingest_expression_correct, test_ingest_expression_raises_InvalidArgument_correcty) and incorrectly (test_ ingest_expression_raises_exceptions_incorrectly). You would Mock the files and set the return file to a fiven value. I can explain this in further detail.

tests/test_ingest.py
eweitz: > if anything you're testing to see that it was called with the correct arguments

That's not the intent.  

This suite does not test that `firestore.Client()` itself gets expected arguments (it wouldn't; the [code doesn't call it with arguments](https://github.com/broadinstitute/scp-ingest-pipeline/blob/63f0ba61e3f30178b9396a2a5c41592d8e3f7193/ingest/ingest_pipeline.py#L62)), it tests that `IngestPipeline.load_expression_data` gets expected arguments.  The tests make assertions about the code slightly upstream of where (and more granularly than) they ideally would.  The reason, as explained in the docstring starting on line 46:

"Ideally we would mock Firestore itself, but existing libraries to mock Firestore (e.g. https://github.com/mdowds/python-mock-firestore) don't implement important Firestore methods like 'batch()'."

> I think you could do something like firestore.Client = Mock() and mock the whole object

`mock_firestore_client` does mock the whole object, albeit superficially.  If we added `firestore.Client = Mock()`, we'd add it to this patch method as recommended [here](https://stackoverflow.com/questions/21982964/how-to-get-the-call-count-using-mock-patch).  However, while that would enable slightly greater coverage, it would also be more much brittle as it would entail testing more even granular implementation details than done now.  The benefit of greater coverage would be exceeded by the cost of lower test maintainability.

Ideally, we would test not just the high-level input to our loading code as done now, but also the high-level output.

For robust value beyond such very thin mocks, we would need to implement `batch()` and any other missing methods in [`python-mock-firestore`](https://github.com/mdowds/python-mock-firestore).  Then we could make assertions about the state of a free, fast, and realistic in-memory Firestore mock in a way that is high-coverage and durable across internal code changes.

tests/test_ingest.py
eweitz: Given that the code creates an instance of `firestore.Client()` and doesn't accept one as a parameter (see [here](https://github.com/broadinstitute/scp-ingest-pipeline/blob/63f0ba61e3f30178b9396a2a5c41592d8e3f7193/ingest/ingest_pipeline.py#L62)), putting `firestore.Client = Mock()` within the patching method would be preferable, as described [here](https://stackoverflow.com/a/8182480/10564415) and [here](https://stackoverflow.com/a/46865495/10564415).

ingest/ingest_pipeline.py
eweitz: Sounds good, I aim to do this today after demo.

tests/test_ingest.py
eweitz: Great suggestion on testing multiple cases.  That will increase coverage, soundly.  Let's touch base; I plan to push this refinement today.

tests/test_ingest.py
knapii-developments: We should mock all of firestore or firestore.Client when we utilize their functions regardless of what suite we're using. When you mock the client object any function you call from it returns a mocked object. This is because when you mock firestore.Client, all functions of the Client object become mocked because it's implied.

tests/test_ingest.py
eweitz: 155b088 increases coverage per this comment.

eweitz:
Thanks for the insightful review and great discussion, Eno.  Your suggestions yielded concrete improvements.  Looking forward to our broader discussion on testing strategy!

PR #12, opened 2019-07-23
Ingest clusters

knapii-developments:
This has to deal with ticket [SCP-1703](https://broadworkbench.atlassian.net/browse/SCP-1703
). Ticket also provides timing data for 1k, 10k, 100k cells.This commit allows cluster files to be ingested into Firestore. CSV, TXT, and TSV files can be ingested. Testing files for cluster and metadata are included as well. 

Command used: python ingest_pipeline.py ingest_cluster --cluster-file ../tests/data/AB_toy_data_28k_genes_1k_cells.cluster.txt


knapii-developments:
> Please ease code review by setting the base branch to something with more isolated changes, e.g. `ea-ingest-cell-metadata` or `master` after merging that branch.

This has been done.

ingest/clusters.py
```diff
@@ -0,0 +1,80 @@
...
+    def transform(self, rows):
+        """ Add data from cluster files into annotation subdocs in cluster data model"""
+        for idx, column in enumerate(rows):
+            if idx != 0:
+                if self.metadata_types[idx].lower() == 'numeric':
+                    column = round(float(column), 3)
+                elif self.metadata_types[idx].lower() == 'group':
+                    if column not in self.unique_values:
+                        self.unique_values.append(column)
+            annotation = self.header[idx].lower()
```
eweitz: I understand why we `self.metadata_types[idx].lower()`, but this use of `lower()` seems like it would unexpectedly change user data -- e.g. cast "GABAergic cells" to "gabaergic cells".  So I'd suggest:

```suggestion
            annotation = self.header[idx]
```

ingest/clusters.py
```diff
@@ -0,0 +1,80 @@
...
+        self.annotation_subdocs = self.create_annotation_subdocs()
+        # sample size needs to be smaller than amount of points
+        # for each cell annotations if there are any
+
+    def update_points(self):
+        self.top_level_doc['points'] = self.amount_of_lines
+
+    def transform(self, rows):
+        """ Add data from cluster files into annotation subdocs in cluster data model"""
+        for idx, column in enumerate(rows):
```
eweitz: Could you add a brief comment in the code explaining what happens when `idx == 0`?  

I suspect your explanation from https://github.com/broadinstitute/scp-ingest-pipeline/pull/13#discussion_r309249856 applies here, but I won't remember that for long.

ingest/ingest_pipeline.py
```diff
@@ -172,27 +182,23 @@ def ingest_expression(self) -> None:
...
-#         None
-#     """
-#     for data in self.cluster.extract():
-#         self.cluster.transform(data)
-#     self.load_cluster_files()
+    def ingest_cluster(self):
+        """Ingests cluster files into Firestore."""
+        while True:
+            row = self.cluster.extract()
+            if(row == None):
```
eweitz:
```suggestion
            if row == None:
```

ingest/clusters.py
bistline: This is the types header row, not the top one (e.g. 2nd line).  We definitely want to cast these to lower case, as was discovered in [SCP-1729](https://broadworkbench.atlassian.net/browse/SCP-1729).  If we don't, and someone supplies a type header value of 'Group' or 'Numeric', it causes annotation values to be handled incorrectly.

ingest/clusters.py
```diff
@@ -0,0 +1,80 @@
...
+            annotation_value = copy.copy(
+                self.annotation_subdocs[annotation]['value'])
+            annotation_value.append(column)
+            self.annotation_subdocs[annotation]['value'] = annotation_value
+
+    def create_annotation_subdocs(self):
+        """Creates annotation_subdocs"""
+        annotation_subdocs = {}
+        for value in self.header:
+            value = value.lower()
```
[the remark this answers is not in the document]
bistline: Again, this is actually what we want.  Name is a bit of a misnomer here - this is actually the column header value, but in the case of the first 3 (name, x, y, and optionally z) we want to downcase these again so that when we load them for visualization, we don't have to worry about case sensitivity in the query.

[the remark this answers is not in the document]
bistline: So Eric is actually correct - we don't want to case convert this value because it will change the text for an annotation column value, and we don't want that.  So we should only do the case change on comparison:

`value.lower() == 'name':`

However, in the case of x/y/z, we should also downcase what is passed through:

```
elif value.lower() in ('x', 'y', 'z'):
    annotation_subdocs[value] = self.create_metadata_subdoc(value.lower(), 'coordinates')
```

ingest/clusters.py
```diff
@@ -0,0 +1,80 @@
...
+            # perform a shallow copy
+            annotation_value = copy.copy(
+                self.annotation_subdocs[annotation]['value'])
+            annotation_value.append(column)
+            self.annotation_subdocs[annotation]['value'] = annotation_value
+
+    def create_annotation_subdocs(self):
+        """Creates annotation_subdocs"""
+        annotation_subdocs = {}
+        for value in self.header:
```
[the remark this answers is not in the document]
knapii-developments: I thought names had to do with the names of cells in the column and not the names of the headers.

ingest/clusters.py
bistline: I retract this - I was looking at ``self.metadata_types[idx]``, and not ``annotation = self.header[idx].lower()``.  Eric is correct - we do not want to downcast the value of the field header.  We want to preserve that case.

knapii-developments:
> Just changes for the case handling in create_annotation_subdocs() - everything else looks fine to me.

Resolved in 60db95d

ingest/clusters.py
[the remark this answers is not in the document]
knapii-developments: Resolved in 60db95d

ingest/clusters.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def create_metadata_subdoc(self, name, header_value_type, *, value=[], subsampled_annotation=None):
+        """returns metadata subdoc"""
+        return {
+            'name': name,
+            'array_index': 0,
+            'value': value,
+            'array_type': header_value_type,
+            'subsampled_annotation': subsampled_annotation,
+            'subsamp_threashold': "",
```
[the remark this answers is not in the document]
knapii-developments: Resolved in 60db95d

ingest/clusters.py
```diff
@@ -0,0 +1,80 @@
...
+        return annotation_subdocs
+
+    def create_metadata_subdoc(self, name, header_value_type, *, value=[], subsampled_annotation=None):
+        """returns metadata subdoc"""
+        return {
+            'name': name,
+            'array_index': 0,
+            'value': value,
+            'array_type': header_value_type,
+            'subsampled_annotation': subsampled_annotation,
```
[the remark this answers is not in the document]
knapii-developments: Resolved in 60db95d

ingest/clusters.py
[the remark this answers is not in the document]
knapii-developments: Resolved in 60db95d

ingest/gene_data_model.py
```diff
@@ -135,13 +135,17 @@ def chunk_gene_expression_documents(self):
         """
         # sum starts at 59 (because key values take up 59 bytes) plus the
         # storage size of the source file name and file type
+        size_of_cell_names_field = 10 + 1
```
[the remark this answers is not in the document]
knapii-developments: Resolved in 60db95d

ingest/clusters.py
knapii-developments: Resolved in 60db95d

ingest/clusters.py
knapii-developments: Resolved in 60db95d

ingest/ingest_pipeline.py
knapii-developments: Resolved in 311a512

ingest/clusters.py
```diff
@@ -0,0 +1,87 @@
...
+            annot_name = value.casefold()
+            if annot_name == 'name':
+                annotation_subdocs[value] = self.create_metadata_subdoc(
+                    'text', 'cells')
+            elif value in ('x', 'y', 'z'):
+                annotation_subdocs[value] = self.create_metadata_subdoc(
+                    annot_name, 'coordinates')
+            else:
+                annotation_subdocs[value] = self.create_metadata_subdoc(
+                    annot_name, 'annotations')
```
bistline: this still isn't correct - it would result in the same issue @eweitz noted earlier.  If we had an annotation header value of "Cell Type", this would get saved in the sub document as "cell type".  The else should read like this:

``
else:
    annotation_subdocs[value] = self.create_metadata_subdoc(value, 'annotations')
``

eweitz: Resolved in d64a62e.

PR #13, opened 2019-07-29
Ingest Cell Metadata

knapii-developments:
This PR relates to ticket [SCP-1703](https://broadworkbench.atlassian.net/browse/SCP-1703) and allows cell metadata files to be ingested into Firestore. It should be noted that indexing should be turned off before ingesting to avoid indexing errors. There are 2 new files added:

**Cell_Metadata.py**
This class includes methods that allow cell metadata files to be ingested. TSV, text, and CSV file formats are compatible. A developer does not need to be concerned with what file formats are being ingested. Returned data structures are consistent between file formats. 

**ingest_files.py**
This class includes methods for opening and extracting from TSV, CSV, and text file formats. 

Command for ingesting cell metadata into Firestore
python ingest_service.py ingest_cell_metadata --cell-metadata-file ../tests/data/10k_cells_29k_genes.metadata.tsv


eweitz:
CircleCI reports a build failure that looks like a true positive:

```
_______________________ ERROR collecting test_ingest.py ________________________
ImportError while importing test module '/home/circleci/scp-ingest-pipeline/tests/test_ingest.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
test_ingest.py:43: in <module>
    from ingest_pipeline import *
../ingest/ingest_pipeline.py:35: in <module>
    from clusters import Clusters
E   ModuleNotFoundError: No module named 'clusters'
```

See https://circleci.com/gh/broadinstitute/scp-ingest-pipeline/30.

I suspect commenting out `from clusters import Clusters` in `/ingest/ingest_pipeline.py`, along with any related draft code for cluster ingest, would ensure the build passes.  To my understanding, you intend to merge cluster code in a separate PR, #12.

Let me know if I can help getting the build to pass.  Once it does, I'll continue reviewing!

knapii-developments:
Eric-

I saw this yesterday before I left. I agree, it is a true positive and it looks like an easy fix.  I’m writing the draft PR as we speak for the Clusters file. I’ll resubmit and see how build runs. If I need help I’ll make sure ping you!

-Eno

ingest/ingest_files.py
```diff
@@ -0,0 +1,120 @@
...
+        file_type_extract_fns = {
+            'text/csv': self.extract_csv_or_tsv,
+            'text/plain': self.extract_txt,
+            'text/tab-separated-values': self.extract_csv_or_tsv,
+        }
+        return file_type_extract_fns.get(self.file_type)()
+
+    def split_line(self, line):
+        """Splits lines on spaces, tabs, or commas"""
+        return re.findall(r'[^,\s\t]+', line)
```
bistline: Same concern as before - I think we want to be more explicit when we splitting on spaces (like for sparse matrices).

knapii-developments: what?

knapii-developments: Resolved in commit 6f30c9d

ingest/ingest_files.py
```diff
@@ -0,0 +1,126 @@
...
+            next_row = self.file.readline()
+
+            if not next_row:
+                break
+            self.amount_of_lines += 1
+            # Create array with no new line, commas, or tab characters
+            next_row_revised = self.split_line(next_row.replace('\n', ''))
+            return next_row_revised
+
+    def get_next_line(self, *, increase_line_count=True, split_line=True):
```
eweitz: This is the first time I recall seeing a [bare asterisk in function arguments](https://stackoverflow.com/a/14298976/10564415).  Neat use of this new Python 3 feature to force the caller to use named arguments!

ingest/ingest_pipeline.py
```diff
@@ -7,11 +7,18 @@
 
 PREREQUISITES
 You must have Google Cloud Firestore installed, authenticated, and
-configured. Must have Python 3.6 or higher.
+configured. Must have python 3.6 or higher. Indexing must be turned off for
```
eweitz:
```suggestion
configured. Must have Python 3.6 or higher. Indexing must be turned off for
```

ingest/ingest_pipeline.py
```diff
@@ -7,11 +7,18 @@
...
 # Takes expression file and stores it into Firestore
 
+# Ingest cluster file
+python ingest_pipeline.py ingest_cluster --cluster-file ../tests/data/AB_toy_data_portal.cluster.txt
+
+# Ingest Cell Metadata file
+python ingest_pipeline.py ingest_cell_metadata --cell-metadata-file ../tests/data/10k_cells_29k_genes.metadata.tsv
+
 # Ingest dense file
 $python ingest_pipeline.py ingest_expression --matrix-file ../tests/data/dense_matrix_19_genes_100k_cells.txt --matrix-file-type dense
```
eweitz: The prepended `$` has somewhat nagged me, because it requires extra care when copying and pasting.  So I'm glad you omitted it in your most recent example commands.  (Also, thank you for including example commands :).)

For consistency could you please remove `$` from the old examples, or add it to the new ones?

ingest/ingest_pipeline.py
```diff
@@ -125,16 +134,33 @@ def load_expression_data(self, list_of_expression_models: List[Gene]) -> None:
 
                     batch.commit()
 
+    def load_cell_metadata(self):
+        """Loads cell metadata files into firestore."""
```
eweitz:
```suggestion
        """Loads cell metadata files into Firestore."""
```

ingest/ingest_pipeline.py
```diff
@@ -125,16 +134,33 @@ def load_expression_data(self, list_of_expression_models: List[Gene]) -> None:
...
+        for annotation in self.cell_metadata.top_level_doc.keys():
+            doc_ref = self.db.collection(collection_name).document()
+            doc_ref.set(self.cell_metadata.top_level_doc[annotation])
+            try:
+                subcollection_doc = self.cell_metadata.data_subcollection[annotation]
+                doc_ref_sub = doc_ref.collection(subcollection_name).document()
+                doc_ref_sub.set(subcollection_doc)
+            except exceptions.InvalidArgument as e:
+                # Catches invalid argument exception, which error "Maximum
+                # document size falls under
```
eweitz:
```suggestion
                # document size" falls under
```

ingest/ingest_pipeline.py
```diff
@@ -161,10 +211,10 @@ def create_parser():
         description=__doc__,
         formatter_class=argparse.RawDescriptionHelpFormatter)
 
-    subparser = parser.add_subparsers()
+    subargs = parser.add_subparsers()
```
eweitz: `subargs` is a bit of a misnomer; `add_subparsers` returns an ArgumentParser object.


[Convention from the Python docs](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.add_subparsers) would have us name this variable `subparsers`, not `subargs`.

```suggestion
    subparsers = parser.add_subparsers()
```

(And similarly for other references to `subargs`.)

ingest/ingest_pipeline.py
```diff
@@ -207,12 +275,14 @@ def validate_arguments(parsed_args):
     Returns:
         None
     """
-    if parsed_args.matrix_file_type == 'mtx' and (parsed_args.gene_file == None
-                                                    or parsed_args.barcode_file == None):
-        raise ValueError(
-            ' Missing arguments: --gene-file and --barcode-file. Mtx files '
-            'must include .genes.tsv, and .barcodes.tsv files. See --help for '
-            'more information')
+    if 'matrix_file' in parsed_args:
```
eweitz: Having two `if` statements is unnecessary; the second `if` could simply be an `and`.

ingest/cell_metadata.py
```diff
@@ -1,40 +1,99 @@
...
+        self.top_level_doc = self.create_documents(
+            file_path, file_id, study_accession)
+        self.data_subcollection = self.create_subdocuments()
         self.errors = defaultdict(list)
         self.ontology = defaultdict(lambda: defaultdict(set))
         self.type = defaultdict(list)
 
+    def transform(self, row: List[str]) -> None:
+        """ Add data from cell metadata files into data model"""
+        for idx, column in enumerate(row):
```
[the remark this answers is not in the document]
knapii-developments: In this case we want to start at 0. If it's index 0 we want to start that value in self.cell_names.

ingest/ingest_pipeline.py
knapii-developments: It'd through an error when it checks for parsed_args.matrix_file_type if there was an 'and' here. It'd say something like parsed_args has no attribute type 'matrix_file_type'.

eweitz: I don't think it would.

If the proposed `and` were there and `'matrix_file' in parsed_args` evaluated to false, then the conditional expression would immediately return false without evaluating `parsed_args.matrix_file_type`.

ingest/cell_metadata.py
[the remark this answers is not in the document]
eweitz: Ah!  Good point.  It'd be possible to account for that like so:

```
# First row is a cell name
self.cell_names.append(rows[0])

# Subsequent rows are annotations
for idx, column in enumerate(row[1:]):
```

That seems slightly clearer (and shorter), but I won't complain if you keep the current approach :).

ingest/cell_metadata.py
```diff
@@ -1,40 +1,99 @@
...
-        self.file = open(file_path, 'r')
-        self.headers = self.file.readline().rstrip('\n').split('\t')
-        self.metadata_types = self.file.readline().rstrip('\n').split('\t')
+        IngestFiles.__init__(self, file_path, self.ALLOWED_FILE_TYPES)
+        self.headers = self.get_next_line(
+            increase_line_count=False)
+        self.metadata_types = self.get_next_line(
+            increase_line_count=False)
+        # unique values for group-based annotations
+        self.uniqueValues = []
```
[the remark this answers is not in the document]
knapii-developments: Resolved in 00ce5f8

ingest/cell_metadata.py
```diff
@@ -1,40 +1,99 @@
...
+    def transform(self, row: List[str]) -> None:
+        """ Add data from cell metadata files into data model"""
+        for idx, column in enumerate(row):
+            if idx != 0:
+                # if annotation is numeric convert from string to float
+                if self.metadata_types[idx].lower() == 'numeric':
+                    column = round(float(column), 3)
+                elif self.metadata_types[idx].lower() == 'group':
+                    # Check for unique values
+                    if column not in self.uniqueValues:
```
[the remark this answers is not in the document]
knapii-developments: Resolved in 00ce5f8

ingest/cell_metadata.py
```diff
@@ -1,40 +1,99 @@
...
+                        self.uniqueValues.append(column)
```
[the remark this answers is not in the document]
knapii-developments: Resolved in 00ce5f8

ingest/ingest_pipeline.py
knapii-developments: Resolved in 00ce5f8

ingest/ingest_pipeline.py
knapii-developments: Resolved in 00ce5f8

ingest/ingest_pipeline.py
knapii-developments: Resolved in 00ce5f8

ingest/ingest_pipeline.py
knapii-developments: resolved in b4f5c8b

ingest/cell_metadata.py
```diff
@@ -1,40 +1,99 @@
...
+                # Get annotation name from header
+                annotation = self.headers[idx]
+                self.data_subcollection[annotation]['values'].append(column)
+            else:
+                # If column isn't a annotation value, it's a cell name
+                self.cell_names.append(column)
+
+    def create_documents(self, file_path, file_id, study_accession):
+        """Creats top level documents for Cell Metadata data structure"""
```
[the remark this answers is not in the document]
knapii-developments: Resolved in 00ce5f8

ingest/ingest_pipeline.py
knapii-developments: Resolved in 5a5e6d0

ingest/cell_metadata.py
```diff
@@ -103,13 +162,13 @@ def validate_type_annotations(self):
...
 
-    def validate_against_header_count(self, list):
+    def validate_against_header_count(self):
         """Metadata header and type counts should match.
 
         :return: boolean   True if valid, False otherwise
         """
         valid = False
-        if not len(self.headers) == len(list):
+        if not len(self.headers) == len(self.metadata_types):
```
jlchang: thanks for catching that I did not refactor properly when I added metadata_types to the CellMetadata object!

ingest/ingest_files.py
```diff
@@ -0,0 +1,125 @@
...
+        }
+        return file_type_extract_fns.get(self.file_type)()
+
+    def split_line(self, line):
+        """Splits lines on file format-appropriate delimiters"""
+
+        if self.is_MTX:
+            return re.findall(r'[^,\t]+', line)
+        else:
+            return re.findall(r'[^,\s\t]+', line)
```
jlchang: Concerned that splitting on whitespace will cause issues where the ingest file has a .txt suffix but is actually a tab-delimited file where the value between the tab has spaces. example would be "BM_tsne.txt" file from SCP101 (A Census of Immune Cells) where values under the header "CELL_TYPE" looked like:
monocyte CD14+
developing B cells
CD4 T cells

ingest/ingest_files.py
```diff
@@ -0,0 +1,125 @@
...
+            'text/csv': self.extract_csv_or_tsv,
+            'text/plain': self.extract_txt,
+            'text/tab-separated-values': self.extract_csv_or_tsv,
+        }
+        return file_type_extract_fns.get(self.file_type)()
+
+    def split_line(self, line):
+        """Splits lines on file format-appropriate delimiters"""
+
+        if self.is_MTX:
```
jlchang: Unclear to me why this distinction is needed. MTX format has tsv files that shouldn't have spaces or commas but the broader re.readall should be fine for MTX.

knapii-developments: Resolved in 00ce5f8

ingest/ingest_files.py
knapii-developments: Resolved in 00ce5f8

jlchang: I'm not sure I see how commit 00ce5f8 would ensure that this example cluster file is parsed correctly (using "cat -vet" to renter tabs as ^I and end of line as $ so that spaces can be visualized)
NAME^IX^IY^ICLUSTER^ICELL_TYPE$
TYPE^Inumeric^Inumeric^Igroup^Igroup$
MantonBM1_HiSeq_1_AAACCTGAGCAGGTCA-1^I-1.8670639929757706^I2.9621075639874057^I4^ICD4 T cells$
MantonBM1_HiSeq_1_AAACCTGCACACTGCG-1^I1.0457630874000223^I12.07025739768945^I1^Imonocyte CD14+$

it seems the current code would end up with a list of 7 elements for first non-header line and a list of 6 elements for the second non-header line. If the file were properly tagged as a tsv file, each line would have five elements.

bistline: I think this actually is backwards - if the file is MTX, then we _do_ want to allow splitting on space.  We have 10X MTX files that are space delimited.  Otherwise, we should not be splitting on spaces as we know we will be getting cell metadata that has values with spaces in it (e.g. ontology labels).

So I think it should be this instead:

```
if self.is_MTX:
            return re.findall(r'[^,\s\t]+', line)
        else:
            return re.findall(r'[^,\t]+', line
```)

jlchang: If the CLUSTER and CELL_TYPE columns were swapped, would we lose the CLUSTER annotation when the file is parsed as txt instead of tsv?

knapii-developments: No we wouldn't. A txt file is still comma or tab delimited.

PR #14, opened 2019-07-30
Fix Gene model chunking

knapii-developments:
This PR relates to ticket [SCP-1766](https://broadworkbench.atlassian.net/browse/SCP-1766). Prior to this PR, calculating document size to fit the Firestore document limit size was unknown. Now, we can successfully partition documents to fit within the 1 MiB (1,048,576 bytes) constraints. 

Command:
python ingest_pipeline.py ingest_expression --matrix-file ../tests/data/dense_matrix_19_genes_100k_cells.txt --matrix-file-type dense


ingest/cell_metadata.py
```diff
@@ -1,39 +1,94 @@
...
+            file_path, file_id, study_accession)
+        self.data_subcollection = self.create_subdocuments()
+        self.errors = {}
+
+    def transform(self, row: List[str]) -> None:
+        """ Add data from cell metadata files into data model"""
+        for idx, column in enumerate(row):
+            if idx != 0:
+                # if annotation is numeric convert from string to float
+                if self.metadata_types[idx].lower() == 'numeric':
```
bistline: Thank you for calling lower() on the header value!  I'm glad we're learning from previous regressions.

ingest/cell_metadata.py
```diff
@@ -1,39 +1,94 @@
...
+        """Creats top level documents for Cell Metadata data structure"""
+        documents = {}
+
+        # Each annotation value has a top level document
+        for value in self.headers[1:]:
+            # Copy document model so memory references are different
+            copy_of_doc_model = copy.copy({
+                'name': value,
+                'study_accession': study_accession,
+                'source_file_name': file_path.strip("."),
```
bistline: This is fine for now, but now that we've added the file_id to all Gene/Cluster/Cell Metadata documents, I'm wondering if source_file_name is now obsolete.

ingest/cell_metadata.py
```diff
@@ -1,39 +1,94 @@
...
+                'source_file_type': 'metadata',
```
bistline: source_file_type actually isn't in the [Firestore Data Structures](https://docs.google.com/document/d/1ArIKzmJFUh-mjojGXquCjnJVGycrHfdOrrlAO-TG8cc/edit#heading=h.q45cv6vfiqrg) schema for CellMetadata documents - please remove.

ingest/cell_metadata.py
```diff
@@ -1,39 +1,94 @@
...
+
+    def get_collection_name(self):
+        """Returns collection name"""
+        return 'cell_metadata'
+
+    def get_subcollection_name(self):
+        """Returns sub-collection name"""
+        return 'data'
 
     def validate_header_keyword(self):
```
bistline: Similar to the group/numeric problem we had in the past, we should be case insensitive on these checks for NAME/TYPE.

ingest/gene_data_model.py
```diff
@@ -31,6 +31,7 @@ def __init__(self, name: str, source_file_name: str, source_file_type: str, *,
 
         self.name = name.replace('"', '')
         self.gene_id = gene_id
+        print (source_file_name)
```
bistline: Do we still want this print statement here?

ingest/gene_data_model.py
```diff
@@ -124,7 +125,7 @@ def set_expression_scores_and_cell_names(self, expression_scores, cell_names):
 
     def chunk_gene_expression_documents(self):
         """Partitions gene expression documents in storage sizes that are
-            less than 104,857 bytes.
+            less than 1,048,576 bytes.
```
bistline: So glad this works now!

ingest/ingest_files.py
```diff
@@ -0,0 +1,120 @@
...
+        file_type_extract_fns = {
+            'text/csv': self.extract_csv_or_tsv,
+            'text/plain': self.extract_txt,
+            'text/tab-separated-values': self.extract_csv_or_tsv,
+        }
+        return file_type_extract_fns.get(self.file_type)()
+
+    def split_line(self, line):
+        """Splits lines on spaces, tabs, or commas"""
+        return re.findall(r'[^,\s\t]+', line)
```
bistline: I'm concerned about splitting on tab, comma, and space here.  We will have instances (especially with cell metadata files) where there may be a value that is a short phrase that we will want to preserve, and if we split on space then this will break.

knapii-developments:
I can run it. Chunking could be a apart of the reason ingest is taking over 6 hours. But I doubt it’s the sole issue that’s causing ingest to take 6 hours.

ingest/gene_data_model.py
```diff
@@ -137,26 +110,33 @@ def chunk_gene_expression_documents(self):
...
             # This and other storage size calculation figures are derived from:
             # https://cloud.google.com/firestore/docs/storage-size
-            if (sum - 1 - 32) > DOCUMENT_LIMIT_BYTES:
-                end_index = index - 1
+            if (sum + 32) > DOCUMENT_LIMIT_BYTES or index == (len(self.cell_names) - 1):
+                if index == (len(self.cell_names) - 1):
+                    end_index = index
+                else:
+                    end_index = index - 1
+                print(sum)
```
bistline: Unnecessary print statement here.

ingest/ingest_files.py
```diff
@@ -53,7 +53,11 @@ def extract(self):
         return file_type_extract_fns.get(self.file_type)()
 
     def split_line(self, line):
+<<<<<<< HEAD
```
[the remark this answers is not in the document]
knapii-developments: This doesn't actually exist. Check it [out](https://github.com/broadinstitute/scp-ingest-pipeline/blob/5c80e521928a16fcb4d4bd7f36ea2692b109bb2a/ingest/ingest_files.py#L112)

ingest/gene_data_model.py
```diff
@@ -133,26 +134,45 @@ def chunk_gene_expression_documents(self):
...
+                            'expression_scores':  self.expression_scores,
+                            'source_file_name': source_file_name,
+                            'source_file_type': source_file_type,
+                            }
+
+        # starting sum is equal to 34 (because key names take up 59 bytes) plus the
         # storage size of the source file name and file type
-        sum = 59 + len(self.source_file_name) + len(self.source_file_type)
+        size_of_document_name = len(document_name) + 1
+        size_of_cell_names_field = 10 + 1
```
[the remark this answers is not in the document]
knapii-developments: Fear not! It's still there. Take a [look](https://github.com/broadinstitute/scp-ingest-pipeline/blob/5c80e521928a16fcb4d4bd7f36ea2692b109bb2a/ingest/gene_data_model.py#L111)

ingest/gene_data_model.py
knapii-developments: I just have it there for debugging, making sure that it's chunking properly, and for other developers. It's easy to comment out if you're uninterested instead of figuring out where to print the sum.

PR #30, opened 2019-09-23
Add tests for ontology validation

jlchang:
This PR is intended to unblock [SCP-1844](https://broadworkbench.atlassian.net/browse/SCP-1844) by adding valid and invalid metadata files that conform to the latest metadata convention for Alexandria (AMC_v1.1.00.json)

Included in this PR:
• support for HCA-style ontologyID syntax in metadata validation
• fixed inaccurate counting of ontology errors
• removed superfluous serialization of CellMetadata.issues in tests
• collection of warnings (formerly printed at the time issue was encountered) into CellMetadata.issues for reporting in report_issues function


ingest/validation/validate_metadata.py
```diff
@@ -234,39 +235,39 @@ def process_metadata_row(metadata, convention, line):
         if k in metadata.type['convention']['integer']:
             try:
                 row_info[k] = int(v)
-            except ValueError as e:
-                error_msg = f'ERROR: value provided does not match convention type delaration for {k}'
+            except ValueError:
```
eweitz: Exception handling looks identical across all three cases here.

Moving the `try` just before and the `except` just after the `if k...` block would eliminate two extraneous `try`/`except` blocks.

ingest/validation/validate_metadata.py
```diff
@@ -234,39 +235,39 @@ def process_metadata_row(metadata, convention, line):
...
-                    'error', 'type', error_msg, [row_info['CellID'], repr(e)]
+                    'error', 'type', error_msg, [row_info['CellID']]
                 )
     return row_info
 
 
-def process_metadata_content(metadata, convention):
-    """Evaluate TSV metadata input non-ontology errors and ontology info
+def collect_jsonschema_errors(metadata, convention):
+    """Evaluate TSV metadata input against metadata convention using jsonschema
```
```suggestion
    """Evaluate TSV metadata input against metadata convention using JSON Schema
```

ingest/validation/validate_metadata.py
```diff
@@ -342,15 +359,19 @@ def retrieve_ontology_term(convention_url, ontology_id):
...
+    # separate ontology shortname from term ID number
+    # valid separators are underscore and colon (used by HCA)
     try:
-        ontology_shortname, term_id = ontology_id.split('_')
+        ontology_shortname, term_id = re.split('[_:]', ontology_id)
+    # when ontolgyID is malformed and has no separator -> ValueError
     except ValueError as error:
         print("ValueError:", error)
-        print('missing ontology shortname in', ontology_id)
+        print('Problem with provided ontologyID', ontology_id)
```
eweitz: If "ontologyID" is a term the user would see, I'd suggest adding a space to make it read more conventionally -- i.e.:
```suggestion
        print('Problem with provided ontology ID', ontology_id)
```

And similarly for other user-facing instances of the term.

tests/data/metadata_invalid.json
```diff
@@ -1 +1 @@
...
\ No newline at end of file
+{"error": {"format": {"Error:  Duplicate CellID(s) in metadata file": ["BM01_16dpp_AAGCAGTGGTAT"]}, "type": {"ERROR: value provided does not match convention type declaration for organism_age": ["BM01_16dpp_CGGTAAACCATT"]}, "convention": {"'ethnicity' is a dependency of 'ethnicity__ontology_label'": ["BM01_16dpp_AAGCAGTGGTAT", "BM01_16dpp_TAAGCAGTGGTA", "BM01_16dpp_CTAAGCAGTGGT", "BM01_16dpp_CGGTAAACCATT", "BM01_16dpp_AAGCAGTGGTAT"], "'sex' is a required property": ["BM01_16dpp_AAGCAGTGGTAT", "BM01_16dpp_TAAGCAGTGGTA", "BM01_16dpp_CTAAGCAGTGGT", "BM01_16dpp_CGGTAAACCATT", "BM01_16dpp_AAGCAGTGGTAT"], "None is not one of ['yes', 'no', 'unknown']": ["BM01_16dpp_TAAGCAGTGGTA"], "None is not of type 'string'": ["BM01_16dpp_TAAGCAGTGGTA"], "'direct from donr - fresh' is not one of ['cell line', 'organoid', 'direct from donor - fresh', 'direct from donor - frozen', 'cultured primary cells']": ["BM01_16dpp_CTAAGCAGTGGT"], "'foo' is not of type 'number'": ["BM01_16dpp_CGGTAAACCATT"]}}, "warn": {"ontology": {"Warning: no ontology label supplied in metdata file for \"UO_0000036\" - no cross-check for data entry error possible": null}}}
```
eweitz: In "Warning: no ontology label supplied in metdata file for", the term "metdata" should be "metadata".

It looks like the proper place to fix that would be upstream, which would then be used to generate new test data with the typo fix.

tests/test_validate_metadata.py
```diff
@@ -1,3 +1,22 @@
+"""Tests for metadata validation
+
+These tests verify that metadata files are checked against metadata convention,
+ontology terms are validated against an external source, and tsv metadata files
+conform to SCP metadata file format requirements.
+
+PREREQUISITEs
```
eweitz:
```suggestion
PREREQUISITES
```

tests/test_validate_metadata.py
```diff
@@ -103,15 +119,13 @@ def test_valid_nonontology_content(self):
     def test_invalid_nonontology_content(self):
         """Non-ontology metadata should conform to convention requirements
             """
-        args = "../tests/data/AMC_v0.8.json " "../tests/data/metadata_invalid.tsv"
+        args = "../tests/data/AMC_v1.1.00.json " "../tests/data/metadata_invalid.tsv"
```
eweitz: The double 0 padding in `AMC_v1.1.00.json` is somewhat unconventional.  [Semantic Versioning](https://semver.org) uses unpadded integers for each of the major, minor and patch versions -- e.g. `AMC_v1.1.0.json` would be more conventional versioning.

tests/test_validate_metadata.py
```diff
@@ -125,18 +139,57 @@ def test_invalid_nonontology_content(self):
...
             reference_issues,
             "Metadata validation issues do not match reference issues",
         )
 
         self.teardown_metadata(metadata)
 
+    def test_valid_ontology_content(self):
+        """Ontology metadata should conform to convention requirements
+            """
+        args = "../tests/data/AMC_v1.1.00.json " "../tests/data/ontology_valid.tsv"
```
eweitz: Simpler equivalent:

```suggestion
        args = "../tests/data/AMC_v1.1.00.json ../tests/data/ontology_valid.tsv"
```

(I had never seen that syntax before!)

ingest/cell_metadata.py
```diff
@@ -182,11 +182,12 @@ def validate_header_keyword(self):
         if self.headers[0].casefold() == 'NAME'.casefold():
             valid = True
             if self.headers[0] != 'NAME':
-                # ToDO - capture warning below in issue report
-                print(
+                msg = (
                     f'Warning: metadata file keyword NAME provided as '
                     f'{self.headers[0]}'
                 )
+                self.store_validation_issue('warn', 'format', msg, '')
```
eweitz: Given the method signature:

`def store_validation_issue(self, type, category, msg, associated_info=None)`

and looking at the method, it seems

`self.store_validation_issue('warn', 'format', msg, '')`

and

`self.store_validation_issue('warn', 'format', msg)`

would result in the same ultimate output.  So this seems simpler:

```suggestion
                self.store_validation_issue('warn', 'format', msg)
```

And similarly for the other cases of passing `''` in this parameter.

ingest/validation/validate_metadata.py
jlchang: I wondered what I could do to make the code cleaner but didn't see a solution - many thanks for the suggestion!

ingest/cell_metadata.py
jlchang: Thanks for catching that - had trouble getting store_validation_issue to work at one point and I think that was my chasing down a red herring...

ingest/validation/validate_metadata.py
jlchang: updated in `4023a08`

ingest/validation/validate_metadata.py
jlchang: updated in `4023a08`

ingest/validation/validate_metadata.py
jlchang: updated in `4023a08`

tests/data/metadata_invalid.json
jlchang: updated code and reference issues file in `4023a08`

tests/test_validate_metadata.py
jlchang: updated in `4023a08`

tests/test_validate_metadata.py
jlchang: updated in `4023a08`

tests/test_validate_metadata.py
jlchang: fixed multiple instances; see `4023a08`
not sure why I had that syntax...

ingest/cell_metadata.py
jlchang: updated in `4023a08`

ingest/cell_metadata.py
```diff
@@ -182,14 +182,15 @@ def validate_header_keyword(self):
         if self.headers[0].casefold() == 'NAME'.casefold():
```
knapii-developments: I'm curious to know why you use casefold vs lower()?

jlchang: Had read that casefold handles more cases than lower but that's not relevant when we're expecting variants on "NAME" and "TYPE" so I've changed the code for better readability.
Reference about why/when casefold would be more appropriate: https://www.w3.org/International/wiki/Case_folding
code updated in `4563451`

PR #31, opened 2019-09-26
Syntax validation for metadata of class "ontology" using JSON schema

jlchang:
This PR is will update serialize_convention.py to generate metadata conventions which validate “ontology” class metadata, allowing both : and _ as separators between the ontology shortname and the ontology ID number.

This update also:
• fixed bug where some ontology errors were overwritten and not reported
• updates ontology_valid.tsv so the test data includes a metadata item using the : separator in our automated tests
• adds test for serialize_convention.py

This fulfills [SCP-1782](https://broadworkbench.atlassian.net/browse/SCP-1782).


tests/test_serialize_convention.py
```diff
@@ -0,0 +1,62 @@
+"""Tests for serialize_convention
+
+This test verifys that a known good input metadata convention tsv file
```
eweitz:
```suggestion
This test verifies that a known good input metadata convention TSV file
```

PR #33, opened 2019-09-30
Enable distribution of Python package

eweitz:
This enables distribution of SCP Ingest Pipeline as a Python package.  It lets this code be imported in other, external Python codebases.

For example, the [`single_cell_portal`](https://github.com/broadinstitute/single_cell_portal) repo can now specify this repo as a dependency, and import this package in [`manage_study.py`](https://github.com/broadinstitute/single_cell_portal/blob/master/scripts/manage_study.py), our public CLI.

Also, for command-line UI polish, validation errors are now printed red and warnings are yellow.

This and broadinstitute/single_cell_portal/pull/79 satisfy [SCP-1900](https://broadworkbench.atlassian.net/browse/SCP-1900).


ingest/validation/validate_metadata.py
```diff
@@ -24,7 +24,12 @@
 import jsonschema
 
 sys.path.append('..')
-from cell_metadata import CellMetadata
+try:
+    # Used when importing internally and in tests
+    from cell_metadata import CellMetadata
+except ImportError:
+    # Used when importing as external package, e.g. imports in single_cell_portal code
+    from ..cell_metadata import CellMetadata
```
jlchang: is this just unfamiliar syntax to me or should there be a slash in ../cell_metadata?

eweitz: Python relative path syntax differs from that of POSIX.  It seemed odd to me at first, but the lack of a slash here is standard; see e.g. https://stackoverflow.com/a/1054281.

PR #34, opened 2019-10-03
Initiate Subsampling

knapii-developments:
This PR communicates status of subsampling with a return code. It also stores subsampled data into Firestore. Rails will be able to invoke and provide status updates about ingest subsampling.

Assumptions:

- Rails determines when to invoke subsampling. 

- It passes both the file ID and study accession to ingest. 

This PR satisfies [SCP-1827](https://broadworkbench.atlassian.net/browse/SCP-1827)

The QA Design will be addressed in [SCP-1835](https://broadworkbench.atlassian.net/browse/SCP-1835)


knapii-developments:
> Looks good! Per your plan, I'll keep an eye out for the future PR containing tests for this.

I added a comment addressing QA design to reflect this statement.

ingest/clusters.py
```diff
@@ -116,7 +137,7 @@ def create_cluster_subdoc(
             "name": annot_name,
             "array_index": 0,
             "values": values,
-            "array_type": header_value_type,
+            "array_type": annot_type,
```
bistline: Not a blocking concern, but calling this variable `annot_type` is somewhat misleading.  That makes me think of `group` or `numeric`, when in fact this value is either `cells`, `coordinates`, or `annotations`

PR #35, opened 2019-10-09
Update metadata validation tests to use AMC v1.1.1

jlchang:
This PR updates metadata validation tests to use latest Alexandria metadata convention (v1.1.1).

This update also:
- improves error messages by including the metadata header with the error, where possible
- provides wrapper function "validate_input_metadata" instead of calling separate validation steps

This fulfills [SCP-1921](https://broadworkbench.atlassian.net/browse/SCP-1921).


ingest/cell_metadata.py
```diff
@@ -305,10 +307,12 @@ def validate_against_header_count(self):
     def validate_format(self):
         """Check all metadata file format criteria for file validity
         """
-        return (
-            self.validate_header_keyword()
-            and self.validate_type_keyword()
-            and self.validate_type_annotations()
-            and self.validate_unique_header()
-            and self.validate_against_header_count()
+        return all(
```
eweitz: Nice use of [`all`](https://docs.python.org/3/library/functions.html#all)!

ingest/validation/validate_metadata.py
```diff
@@ -180,43 +182,41 @@ def validate_cells_unique(metadata):
         valid = True
     else:
         dups = list_duplicates(metadata.cells)
-        msg = 'Error:  Duplicate CellID(s) in metadata file'
-        metadata.issues['error']['format'][msg] = dups
+        error_msg = 'Error:  Duplicate CellID(s) in metadata file'
```
eweitz: To prevent redundant text ("Error: Error: ") in issue messages:

```suggestion
        error_msg = 'Duplicate CellID(s) in metadata file'
```

tests/test_validate_metadata.py
```diff
@@ -31,17 +31,18 @@
...
 )
 
 
 class TestValidateMetadata(unittest.TestCase):
     def setup_metadata(self, args):
-        args_list = args.split(" ")
+        args_list = args.split(' ')
         args = create_parser().parse_args(args_list)
-        with open(args.convention, "r") as f:
+        with open(args.convention, 'r') as f:
```
eweitz: Read more (`r`) is default in [open()](https://docs.python.org/3/library/functions.html#open), so best to omit like we do for other defaults:

```suggestion
        with open(args.convention) as f:
```

ingest/validation/validate_metadata.py
jlchang: Thanks for the catch!

jlchang: Updated in d58de2c. Reference json updated for affected test.

tests/test_validate_metadata.py
jlchang: Updated in d58de2c.

PR #36, opened 2019-10-23
Validation of array-based metadata (SCP-1868)

jlchang:
This PR implements the validation of array-based metadata by implementing parsing and type casting of array data (delimited by pipe "|" character) for JSON schema validation.

Also included in this PR:
• parsing and type casting of boolean values (when not all values in column are true boolean)
• detect when TYPE annotation disagrees with declared metadata convention type
• valid metadata headers only contain alphanumeric characters and underscore
• check all values for __unit metadata are uniform
• Alexandria metadata v1.1.2 - correction of missing __unit metadata and ARRAY settings
• Alexandria metadata v1.1.3 - set biosample_id to required, add donor_id as required

This PR fulfills [SCP-1868](https://broadworkbench.atlassian.net/browse/SCP-1868)


ingest/validation/validate_metadata.py
```diff
@@ -187,72 +178,181 @@ def validate_cells_unique(metadata):
...
+                'error', 'type', error_msg, [row_info['CellID']]
+            )
+        # particular metadatum is not in convention, value not needed for validation
+        except TypeError:
+            row_info[metadatum] = ''
+    return row_info
+
+
+def compare_type_annots_to_convention(metadata, convention):
+    """Check metadata type annotation consistent with metadata convention type
```
eweitz: Slightly clearer:

```suggestion
    """Check if metadata type annotation is consistent with metadata convention type
```

ingest/validation/validate_metadata.py
```diff
@@ -478,37 +579,50 @@ def validate_collected_ontology_data(metadata, convention):
...
 def serialize_issues(metadata):
     """Write collected issues to json file
     """
     with open('issues.json', 'w') as jsonfile:
-        json.dump(metadata.issues, jsonfile)
+        json.dump(metadata.issues, jsonfile, indent=2)
+
+
+def review_metadata_names(metadata):
+    """check metadata names for disallowed characters
```
```suggestion
    """Check metadata names for disallowed characters
```

ingest/validation/validate_metadata.py
```diff
@@ -187,72 +178,181 @@ def validate_cells_unique(metadata):
...
+    """
+    if value.lower() == 'true':
+        return True
+    elif value.lower() == 'false':
+        return False
+    else:
+        return value
+
+
+def cast_integer_type(value):
```
eweitz: At first glance this struck me as over-engineering, but after reviewing more closely how you use it in e.g. line 269-271 this seems reasonable.

tests/test_validate_metadata.py
```diff
@@ -134,22 +133,19 @@ def test_invalid_nonontology_content(self):
...
-
         self.teardown_metadata(metadata)
 
     def test_valid_ontology_content(self):
         """Ontology metadata should conform to convention requirements
             """
-        args = '../tests/data/AMC_v1.1.1.json ../tests/data/valid_v1.1.1.tsv'
+        # Note: this input metadata file does not have array-based metadata
+        # is compatible with v1.1.2 but not v1.1.3 (missing sampleID and donorID)
+        args = '../tests/data/AMC_v1.1.2.json ../tests/data/valid_v1.1.1.tsv'
```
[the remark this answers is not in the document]
jlchang: "valid_v1.1.1.tsv" is valid for both ontology and metadata cases - the same file is used for both tests. I see that it makes the naming more opaque but the suggested change here would be confusing for the metadata test. Perhaps valid_metadata_ontology_v.1.1.1.tsv?

tests/test_validate_metadata.py
```diff
@@ -190,7 +188,57 @@ def test_invalid_ontology_content(self):
             'Ontology validation issues do not match reference issues',
         )
         reference_file.close()
+        self.teardown_metadata(metadata)
 
+    def test_valid_array_content(self):
+        """array-based metadata should conform to convention requirements
+            """
```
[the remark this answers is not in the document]
jlchang: addressed in 95d0e50

tests/test_serialize_convention.py
```diff
@@ -7,7 +7,7 @@
 Spin up Python 3.6 virtualenv, install Python dependencies in requirements.txt
 
 # Run test
-python3 test_serialize_convention.py
+python3 test_serialize_convention.py -p . Alexandria ../tests/data/AMC_v1.1.2.tsv
```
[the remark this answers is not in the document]
jlchang: fixed in 95d0e50

ingest/validation/validate_metadata.py
jlchang: updated in 95d0e50

ingest/validation/validate_metadata.py
jlchang: updated in 95d0e50

ingest/validation/validate_metadata.py
jlchang: Thanks - originally had both array handling and type casting in same if-elif and realized there were combinations of cases I was missing (and readability was terrible). This approach seemed to cover all cases. If there is a better way, I'm interested to know what other options I should consider next time.

PR #38, opened 2019-10-24
Write QA Test for Subsampling

knapii-developments:
Before this PR, there were no test for subsampling. Now critical functionality is tested such as:

- Groups that have numeric values are being treated as strings
- Numeric columns are rounded to 3 decimals points
- Binning correctly
- Subsampling correctly
- Filtering cell names (given from cluster file) in metadata correctly

In order to test the above cases the file requirements are:
- Has 3D data
- Has both numeric and group annotations
- Cluster file has less cells than metadata file
- Has a group annotation column that has numeric values

This PR satisfies [SCP-1835](https://broadworkbench.atlassian.net/browse/SCP-1835)


tests/test_ingest_files.py
```diff
@@ -0,0 +1,95 @@
...
+
+These tests verifies:
+    - Group type annotations that have numeric-like values are being treated as strings
+    - Numeric columns are rounded to 3 decimals points
+    - Filtering cell names (given from cluster file) in metadata correctly
+
+PREREQUISITES
+Spin up Python 3.6 virtualenv, install Python dependencies in requirements.txt
+and Firestore emulator must be running, see PR26 for instructions
+(https://github.com/broadinstitute/scp-ingest-pipeline/pull/26)
```
eweitz: Is a running Firestore emulator truly a prerequisite here?  These tests seem independent from Firestore.

tests/test_subsample.py
```diff
@@ -0,0 +1,97 @@
+"""Tests subsampling
+
+These tests verifies:
+    - Binning correctly
+    - Subsampling correctly
+
+PREREQUISITES
+Spin up Python 3.6 virtualenv, install Python dependencies in requirements.txt
+and Firestore emulator must be running, see PR26 for instructions
+(https://github.com/broadinstitute/scp-ingest-pipeline/pull/26)
```
eweitz: As in `test_ingest_files`, I suspect Firestore might not be required to run the tests in this file.

tests/test_subsample.py
```diff
@@ -0,0 +1,97 @@
...
+import numpy as np
+
+sys.path.append("../ingest")
+
+from subsample import SubSample
+
+
+class TestSubsample(unittest.TestCase):
+    AMOUNT_OF_NUMERIC_BINS = 20
+    CLUSTER_PATH = '../tests/data/test_1k_cluster_Data.csv'
```
eweitz: Conventional casing:

```suggestion
    CLUSTER_PATH = '../tests/data/test_1k_cluster_data.csv'
```

tests/test_subsample.py
```diff
@@ -0,0 +1,97 @@
+"""Tests subsampling
+
+These tests verifies:
```
eweitz:
```suggestion
These tests verify:
```

tests/test_ingest_files.py
```diff
@@ -0,0 +1,95 @@
+"""Test ingest_files.py
+
+These tests verifies:
```
eweitz:
```suggestion
These tests verify:
```

tests/test_ingest_files.py
```diff
@@ -0,0 +1,95 @@
...
+PREREQUISITES
+Spin up Python 3.6 virtualenv, install Python dependencies in requirements.txt
+and Firestore emulator must be running, see PR26 for instructions
+(https://github.com/broadinstitute/scp-ingest-pipeline/pull/26)
+
+Note: When CI environment moves to Python 3.7, tests may break due to minor
+differences in how the reference issues are serialized
+
+# Run all tests in a manner that shows report_issues output
+python3 test_ingest_files.py -s
```
eweitz: Nice refinement!

ingest/ingest_pipeline.py
```diff
@@ -49,7 +49,11 @@
 from mtx import Mtx
 from subsample import SubSample
 from loom import Loom
-from validation.validate_metadata import validate_input_metadata, report_issues
```
jlchang: validate_input_metadata will call all functions needed for a valid metadata file (including collect_jsonschema_errors and validate_collected_ontology_data). If you only call the two functions and not validate_input_metadata, then full metadata file validation is not being performed. Are there  issues with the functions recently added to validate_input_metadata (review_metadata_names and confirm_uniform_units) that need discussion?

knapii-developments: I think this comment was misplaced and may have to deal with the metadata review and not subsampling. @jlchang is that correct?

tests/test_ingest_files.py
knapii-developments: You're right. Even with copy and paste, reading is essential!

ingest/ingest_pipeline.py
knapii-developments: I've addressed this issue in ddb16a0

tests/test_subsample.py
knapii-developments: 6344fd2

ingest/ingest_pipeline.py
jlchang: Thanks Eno. I was worried we'd silently switch back to an earlier state and not run validation in ingest_pipeline as intended. This update addresses my concerns.

PR #39, opened 2019-10-24
Store metadata convention for Ingest Pipeline to access

knapii-developments:
This PR allows ingest to validate against a metadata convention that is stored in the [google bucket](https://console.cloud.google.com/storage/browser/_details/fc-bcc55e6c-bec3-4b2e-9fb2-5e1526ddfcd2/metadata_conventions/AMC_v1.1.1.json). 

Permissions:
You must have have storage.buckets.get access to the [google bucket](https://console.cloud.google.com/storage/browser/fc-bcc55e6c-bec3-4b2e-9fb2-5e1526ddfcd2/metadata_conventions/) where the metadata convention json file is stored.

Some important assumptions:
- Users are not allowed to provide a metadata convention
- Metadata convention will not change frequently
- Users do not have permissions to update metadata convention file and that responsibility will be with a member of SCP


ingest/ingest_pipeline.py
```diff
@@ -61,7 +63,7 @@
 
 class IngestPipeline(object):
     # File location for metadata json convention
-    JSON_CONVENTION = 'DoNotTouch/AMC_v0.8.json'
+    JSON_CONVENTION = 'gs://fc-bcc55e6c-bec3-4b2e-9fb2-5e1526ddfcd2/metadata_conventions/AMC_v1.1.1.json'
```
jlchang: We should use AMC_v1.1.3
I've added the appropriate files to the bucket:
gs://fc-bcc55e6c-bec3-4b2e-9fb2-5e1526ddfcd2/metadata_conventions/AMC_v1.1.3/AMC_v1.1.3.json
gs://fc-bcc55e6c-bec3-4b2e-9fb2-5e1526ddfcd2/metadata_conventions/AMC_v1.1.3/valid_array_v1.1.3.tsv
I think we still need a passing full metadata file before we'd update to AMC_v1.1.3?

ingest/ingest_pipeline.py
```diff
@@ -232,8 +234,8 @@ def load_subsample(self, doc):
 
     def has_valid_metadata_convention(self):
         """ Determines if cell metadata file follows metadata convention"""
-        with open(self.JSON_CONVENTION, 'r') as f:
-            convention = json.load(f)
+        json_file = IngestFiles(self.JSON_CONVENTION, ['application/json'])
```
jlchang: just my ignorance here, should we need to close this filehandle somewhere? In my tests for validate_metadata, I'd get errors about filehandles. I'm trying to understand best practice around this.

knapii-developments: 0dba75b.

ingest/ingest_pipeline.py
knapii-developments: 0dba75b

PR #40, opened 2019-10-29
Replicate Rails models for clusters

knapii-developments:
This PR provides programming that replicates Rails models for clusters. Please take note that legacy code will still exist until requirements and functionality are present/converted for ingest to run and tests to pass. **This PR just creates the model and does not write it to JSON yet. Data model attribute _linear_data_id_ has not been set as well.** Once all Rail data models have been implemented, outputting JSON will be addressed. It is believed that models will have common attributes and or models. Waiting to write out JSON will help the design process by eliminating possible redundant code. 

New command: 
python ingest_pipeline.py --study-id SCP1 --study-file-id 123abc ingest_cluster --cluster-file ../tests/data/test_1k_cluster_Data.csv --ingest-cluster --name cluster1 --domain-ranges "{'x':[-1, 1], 'y':[-1, 1], 'z':[-1, 1]}"


ingest/clusters.py
```diff
@@ -1,45 +1,77 @@
...
+    y: List
+    z: List = None
 
-    ALLOWED_FILE_TYPES = ["text/csv", "text/plain", "text/tab-separated-values"]
-    COLLECTION_NAME = "clusters"
+
+class Clusters(Annotations):
+    # Need to delete after refactor
+    COLLECTION_NAME = "cluster_groups"
     SUBCOLLECTION_NAME = "data"
```
bistline: We don't need this anymore, just FYI.

ingest/clusters.py
```diff
@@ -1,45 +1,77 @@
...
-        self.header = self.get_next_line(increase_line_count=False)
-        # Second line in cluster is metadata_type
-        self.metadata_types = self.get_next_line(increase_line_count=False)
+        Annotations.__init__(self, file_path, self.ALLOWED_FILE_TYPES)
+        self.determine_coordinates_and_cell_names()
         self.source_file_type = "cluster"
-        self.has_z = True if ("z" in self.header or "Z" in self.header) else False
-        self.cell_annotations = self.create_cell_annotations_field()
+        self.cluster_type = (
+            '2d'
```
bistline: This is backwards - it should be 2d if z/Z is NOT in the headers.

ingest/clusters.py
```diff
@@ -1,45 +1,77 @@
...
+        self.domain_ranges = domain_ranges
+        self.study_id = study_id
+        self.study_file_id = study_file_id
         # TODO: Populate the cell_annotations array when pandas is implemented
         self.top_level_doc = {
             "name": name,
-            "cluster_type": "3d" if self.has_z else "2d",
             "cell_annotations": [],
-            "study_accession": study_accession,
+            "study_accession": study_id,
```
bistline: study_accession and study_id are not the same thing.  The accession is SCP123, whereas the study_id is a MongoDB document id (like 5d276a50421aa9117c982845).

ingest/clusters.py
knapii-developments: 91e7310

ingest/clusters.py
knapii-developments: I know.

ingest/clusters.py
knapii-developments: This is irrelevant right. The code is still here untill I make the final transition.

tests/test_ingest.py
```diff
@@ -3,40 +3,43 @@
 These tests verify that various matrix file types can be extracted and
 transformed, as expected by code that loads transformed data into Firestore.
 
-Test doubles are used for test speed and isolation.
+Mocks are used for test speed and isolation.
```
eweitz: It looks like the fresh content in `master` was overwritten by outdated content in your branch.  

In other words, the upstream version of `test_ingest.py` (left, red) is what was updated last week to account for removing the Firestore emulator (#44), plus Jean's updates testing gzipped files (#45).  The version you've committed here reverts all that and is outdated.

Quick fix: replace all the content in this file with that from `master`, i.e. copy https://raw.githubusercontent.com/broadinstitute/scp-ingest-pipeline/0b6cb898ab9f473e1b654876cdc0f552518e92b1/tests/test_ingest.py and replace this file with that.

ingest/annotations.py
```diff
@@ -0,0 +1,74 @@
...
+
+        self.preproccess()
+
+    @abc.abstractmethod
+    def transform(self):
+        """Returns data model"""
+
+    # This will end up being a class method
+    @abc.abstractmethod
+    def set_dataArray(self):
```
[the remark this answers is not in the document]
knapii-developments: It's referenced as dataArray in [rails code base](https://docs.google.com/document/d/1BuAjf6ioxNaGFkk6-Eop4AXEOU71ecb3GWY8-QNQEd4/edit#heading=h.mgdh2afryi9l), model name, as well as in the [documents](https://docs.google.com/document/d/1BuAjf6ioxNaGFkk6-Eop4AXEOU71ecb3GWY8-QNQEd4/). So I'm just trying to remain consistent.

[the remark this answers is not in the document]
eweitz: I don't see `dataArray` (i.e. lower camel case) anywhere in those links.

Like Python, the Rails codebase uses [`DataArray`](https://github.com/broadinstitute/single_cell_portal_core/blob/master/app/models/data_array.rb#L1) (i.e. upper camel case) for class references and [`data_array`](https://github.com/broadinstitute/single_cell_portal_core/search?q=data_array&unscoped_q=data_array) (i.e. snake case) for object references.  So the suggested change would be consistent with Rails and Python conventions, and conventions in our Rails codebase.

Also, you use "set_data_array" elsewhere in this PR (in commented-out code), so the suggested would also make this code more internally consistent.

ingest/ingest_pipeline.py
```diff
@@ -13,26 +13,26 @@
 # Takes expression file and stores it into Firestore
 
 # Ingest cluster file
-python ingest_pipeline.py --study-accession SCP1 --file-id 123abc ingest_cluster --cluster-file ../tests/data/test_1k_cluster_Data.csv --ingest-cluster --name cluster1 --domain-ranges "{'x':[-1, 1], 'y':[-1, 1], 'z':[-1, 1]}"
+python ingest_pipeline.py --study-id 5d276a50421aa9117c982845 --study-file-id 123abc ingest_cluster --cluster-file ../tests/data/test_1k_cluster_Data.csv --ingest-cluster --name cluster1 --domain-ranges "{'x':[-1, 1], 'y':[-1, 1], 'z':[-1, 1]}"
 
 # Ingest Cell Metadata file
-python ingest_pipeline.py --study-accession SCP1 --file-id 123abc ingest_cell_metadata --cell-metadata-file ../tests/data/valid_v1.1.1.tsv --ingest-cell-metadata
+python ingest_pipeline.py --study-accession 5d276a50421aa9117c982845 --file-id 123abc ingest_cell_metadata --cell-metadata-file ../tests/data/valid_v1.1.1.tsv --ingest-cell-metadata
```
bistline: These should say `--study-id`, and not `--study-accession`.

ingest/annotations.py
```diff
@@ -0,0 +1,76 @@
...
+        group_columns = self.file.xs(
+            "group", axis=1, level=1, drop_level=False
+        ).columns.tolist()
+        self.file[group_columns] = self.file[group_columns].astype(str)
+        # Find numeric columns and round to 3 decimals places and are floats
+        numeric_columns = self.file.xs(
+            "numeric", axis=1, level=1, drop_level=False
+        ).columns.tolist()
+        # TODO perform replace
+        self.file[numeric_columns] = self.file[numeric_columns].round(3).astype(float)
```
jlchang: @knapii-developments can we look at when preprocess is invoked? I seem to recall trying to run ingest pipeline on an invalid file (which would fail validation) and hitting a runtime error because the invalid file has a string that cannot be coerced into a float. I'll try and reproduce the error I'm recalling so we can take a look at the potential issue.

jlchang: Here's the ingest command of invalid data that fails in preprocess at runtime:
 python ingest_pipeline.py --study-accession SCP1 --file-id 123abc ingest_cell_metadata --cell-metadata-file ../tests/data/invalid_metadata_v1.1.1.tsv --ingest-cell-metadata

ValueError: could not convert string to float: 'foo'

PR #43, opened 2019-10-30
generate json files for BigQuery testing [SCP-1922, SCP-1954]

jlchang:
This PR supports testing BigQuery for handling of array-based metadata.

`convention_to_bq_schema.py` generates a json file describing the schema of a BigQuery table based on a metadata convention (SCP-1922)
`validate_metadata.py` updated to output NDJSON suitable to upload array-based metadata to a BigQuery table (SCP-1954)
`upload_to_bq.py` an example python script to upload NDJSON to BigQuery (SCP-1954)

Also included in this PR:
• clarified error reporting to address non-intuitive metadata validation error messages due to conflict of value types (SCP-1868)
• update reference test files to ensure that tests for validating metadata still pass

Note: tried to make TSV-specific code more robust to future ingest of non-TSV metadata but found that current code is too tightly tied to expected input file format to easily implement. Commented locations in `validate_metadata.py` to denote areas we will need to check if/when we work with loom or other file formats.


scripts/scratch_query/upload_to_bq.py
```diff
@@ -0,0 +1,57 @@
+"""Upload NDJSON file to BigQuery
+
+DESCRIPTION
+This CLI takes a local NDJSON file and appends it to an existing bigquery table.
+(reference: https://cloud.google.com/bigquery/docs/loading-data-local#loading_data_from_a_local_data_source and
+https://cloud.google.com/bigquery/docs/loading-data-cloud-storage-json#loading_json_data_into_a_new_table)
+
+EXAMPLE
+$ python upload_to_bq.py ../../tests/data/valid_arrays_v1.1.3_for_bq_v1.json
```
eweitz: This example lacks values for the positional arguments `table_id` and `dataset_id`, which you define below.  I assume they're required, as is conventional for positional arguments, so my guess is this example command fails.

If that's correct, could you fix the example command?

tests/data/issues_array_v1.1.2.json
```diff
@@ -21,10 +21,8 @@
       "disease__treated: 'T' is not of type 'boolean'": [
         "BM01_16dpp_CTAAGCAGTGGT"
       ],
-      "disease__time_since_onset: '36|three|1' is not of type 'array'": [
-        "BM01_16dpp_CGGTAAACCATT"
-      ],
-      "disease__time_since_onset: 'zero' is not of type 'array'": [
+      "'disease__time_since_onset' is a dependency of 'disease__time_since_onset__unit'": [
```
eweitz: "depends on" seems like a simpler equivalent to "is a dependency of":

```suggestion
      "'disease__time_since_onset' depends on 'disease__time_since_onset__unit'": [
```

If you have qualms or think it'd be more than a 30-second change across files, please ignore this suggestion!

scripts/scratch_query/upload_to_bq.py
```diff
@@ -0,0 +1,57 @@
+"""Upload NDJSON file to BigQuery
```
eweitz: Today I learned that [newline-delimited JSON](http://ndjson.org/) (NDJSON) is a thing!

tests/data/issues_array_v1.1.2.json
jlchang: I agree the message is hard to parse. Unfortunately this one is generated by the JSON schema validator so it isn't something we can easily modify :(

scripts/scratch_query/upload_to_bq.py
jlchang: Great catch. I've update the example command in 31c8281

scripts/scratch_query/convention_to_bq_schema.py
```diff
@@ -0,0 +1,108 @@
...
+    parser = argparse.ArgumentParser(
+        description=__doc__, formatter_class=argparse.RawDescriptionHelpFormatter
+    )
+    parser.add_argument('--output-path', '-p', help='Path for output file')
+    parser.add_argument('input_convention', help='Metadata convention tsv file')
+    return parser
+
+
+def process_row_type(type_info):
+    type_map = {
```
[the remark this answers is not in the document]
jlchang: It's specific to writing a BigQuery schema. Our current plan is to have exactly one table for Alexandria metadata (and then, hopefully, migrate to using Data Repo)- so I only see running this script during dev and don't see a need for this type_map anywhere else. Thanks for asking!

PR #52, opened 2019-11-20
Write modeled data from Ingest Pipeline to MongoDB

knapii-developments:
This PR allows models from ingest files to be written to MongoDB.


ingest/annotations.py
```diff
@@ -79,7 +79,7 @@ def preproccess(self):
             "numeric", axis=1, level=1, drop_level=False
         ).columns.tolist()
         # TODO perform replace
-        self.file[numeric_columns] = self.file[numeric_columns].round(3).astype(float)
+        self.file[numeric_columns] = self.file[numeric_columns].round(3).astype(int)
```
bistline: How come we're casting it as an int?  Shouldn't it be a float?

knapii-developments: my bad. Had issues loading int64 and this was apart of the debugging process. Fixed in cbb6a24

ingest/ingest_pipeline.py
```diff
@@ -145,54 +145,64 @@ def close_matrix(self):
...
+        try:
+            linear_id = self.db[collection_name].insert(model)
             for data_array_model in set_data_array_fn(
-                'linear_id', *set_data_array_fn_args, **set_data_array_fn_kwargs
+                linear_id, *set_data_array_fn_args, **set_data_array_fn_kwargs
             ):
-                print(data_array_model)
+
+                documents.append(data_array_model)
+            self.db['data_arrays'].insert_many(documents)
```
bistline: Very nice.  It will be interesting to see how well this scales.

knapii-developments:
> for associations, and I don't know if we can generate them without inserting a document first.

I think once we start doing performance evaluations it might be worthwhile if it turn out to be a bottle neck. Otherwise, this meets the 'get it working' and 'get it right' portions.

bistline:
> > for associations, and I don't know if we can generate them without inserting a document first.
> 
> I think once we start doing performance evaluations it might be worthwhile if it turn out to be a bottle neck. Otherwise, this meets the 'get it working' and 'get it right' portions.

Totally agreed on that point.

PR #54, opened 2019-11-21
Write convention metadata to BigQuery [SCP-1941]

jlchang:
This PR leverages existing code for metadata validation to generate an NDJSON file suitable for upload to BigQuery.

To upload metadata, ingest-pipeline.py expects:
• the `--validate-convention` flag
• a valid BigQuery dataset name for the GCP project in the runtime environment
• a valid BigQuery table in the provided BigQuery dataset

`--validate-convention` without `--bq-dataset` and `--bq-table` will validate without uploading convention metadata to BigQuery.

This PR fulfills [SCP-1941](https://broadworkbench.atlassian.net/browse/SCP-1941)

### DIFF
--- a/ingest/ingest_pipeline.py
+++ b/ingest/ingest_pipeline.py
@@ -48,15 +48,27 @@ from pymongo import MongoClient
 from mtx import Mtx
 from google.cloud import storage
 
-# from ingest_files import IngestFiles
-from subsample import SubSample
-from loom import Loom
-from ingest_files import IngestFiles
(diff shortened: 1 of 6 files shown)


ingest/ingest_pipeline.py
```diff
@@ -467,6 +519,31 @@ def create_parser():
     return parser
 
 
+def bq_dataset_exists(dataset):
```
knapii-developments: I think these functions should go in a different place. It's not blocking

PR #55, opened 2019-11-22
Setting type of study_id and study_file_id to ObjectId as needed

bistline:
In order to preserve associations in Mongo, we need to set all "_id" attributes to be a BSON ObjectId rather than a string.  Results returned from inserts/queries will already be in the correct format, but args passed to ingest on the command line are strings and therefore need to be cast.

This satisfies SCP-2006


bistline:
Addressing subsequent issues with early termination of parsing that causes incomplete DataArray writes.

ingest/dense.py
```diff
@@ -83,12 +83,17 @@ def set_data_array(
...
-            gene_name, linear_data_id, list(cells_and_expression_vals)
-        )
-        return self.set_data_array_gene_expression_values(
-            gene_name, linear_data_id, cells_and_expression_vals.values()
-        )
+        observed_cells = list(cells_and_expression_vals)
+        observed_values = list(cells_and_expression_vals.values())
+
+        # only return significant data (skip if no expression was observed)
+        if (len(observed_values) > 0):
```
eweitz: Conventional Python:

```suggestion
        if len(observed_values) > 0:
```

PR #62, opened 2019-12-17
Fix subsampling

knapii-developments:
Prior to this commit there were issues with subsampling functionality and ingesting cluster files. Mainly:

- There were cases when the amount of subsampled values were not equal to the subsample threshold. Tests have been updated to check for the use case. 

- While ingesting cluster files, if there is not a value for the argument _domain-ranges_, this value is reflected as 'null' in the database instead of an empty dictionary. 

- The _cell_annotations_ field in the Cluster data model was not saving numeric group values as strings.


ingest/ingest_pipeline.py
```diff
@@ -593,7 +593,7 @@ def bq_dataset_exists(dataset):
         bigquery_client.get_dataset(dataset_ref)
         exists = True
     except NotFound:
-        print(f'Dataset {dataset} not found')
+        (f'Dataset {dataset} not found')
```
knapii-developments:
```suggestion
        print(f'Dataset {dataset} not found')
```

PR #71, opened 2020-01-07
Detect duplicate genes and improper headers for dense matrices (SCP-2032)

knapii-developments:
This PR restores parity with SCP 1.0  by detecting duplicate gene names and headers in an expression matrix file. Lastly the keyword "GENE" in the first cell in the header is validated. When any of these use cases are encountered, and ingest gracefully shuts down with an exit status of 1. Tests have been added for these cases.

This ticket satisfies SCP-2032


ingest/dense.py
```diff
@@ -52,16 +75,16 @@ def preprocess(self):
...
-        dtypes.update({cell_name: 'float' for cell_name in header[1:]})
-        self.df = self.open_file(
-            self.file_path,
-            open_as='dataframe',
-            names=header,
-            skiprows=1,
-            dtype=dtypes,
-            # chunksize=100000, Save for when we chunk data
-        )[0]
+            # Set dtype for expression values to floats
```
bistline: Should this be indented?  It looks like this block should always be invoked, but right now it would only get called in the `else` case.

tests/data/dense_expression_matrix.txt
```diff
@@ -0,0 +1 @@
+GENE	CELL_0001	CELL_0002	CELL_0003	CELL_0004	CELL_0005	CELL_0006	CELL_0007	CELL_0008	CELL_0009	CELL_00010	CELL_00011	CELL_00012	CELL_00013	CELL_00014	CELL_00015Itm2a	0	0	0	0	0	0	0	0	0	0	0	0	0	0	0Sergef	0	7.092	0	7.511	6.803	0	0	0	6.783	7.562	0	7.073	6.697	0	5.61
```
bistline: It looks like this matrix is malformatted, but isn't marked as a "bad" matrix (it's only one line, and should be several).

tests/data/expression_matrix_bad_missing_keyword.txt
```diff
@@ -0,0 +1 @@
+GE	CELL_0001	CELL_0002	CELL_0003	CELL_0004	CELL_0005	CELL_0006	CELL_0007	CELL_0008	CELL_0009	CELL_00010	CELL_00011	CELL_00012	CELL_00013	CELL_00014	CELL_00015Itm2a	0	0	0	0	0	0	0	0	0	0	0	0	0	0	0Sergef	0	7.092	0	7.511	6.803	0	0	0	6.783	7.562	0	7.073	6.697	0	5.61
```
bistline: Same thing with this matrix - it's only one line.  The newlines got removed somehow.

tests/data/expression_matrix_non_unique.txt
```diff
@@ -0,0 +1 @@
+GENE	CELL_0001	CELL_0001	CELL_0003	CELL_0004	CELL_0005	CELL_0006	CELL_0007	CELL_0008	CELL_0009	CELL_00010	CELL_00011	CELL_00012	CELL_00013	CELL_00014	CELL_00015Itm2a	0	0	0	0	0	0	0	0	0	0	0	0	0	0	0Sergef	0	7.092	0	7.511	6.803	0	0	0	6.783	7.562	0	7.073	6.697	0	5.61
```
bistline: This one too.

ingest/dense.py
knapii-developments: b41dfe8

tests/data/dense_expression_matrix.txt
knapii-developments: b41dfe8

tests/data/expression_matrix_bad_missing_keyword.txt
knapii-developments: b41dfe8

tests/data/expression_matrix_non_unique.txt
knapii-developments: b41dfe8

ingest/dense.py
```diff
@@ -33,17 +33,40 @@ def __init__(self, file_path, study_file_id, study_id, **kwargs):
...
 
-        self.preprocess()
+    def validate_gene_keyword(self):
+        # File is an R formatted file
+        if (self.header[-1] == '') and (self.header[0].upper() != 'GENE'):
+            pass
+        else:
+            # If not R formatted file then first cell must be 'GENE'
+            if self.header[0].upper() != 'GENE':
+                self.error_logger.error(
```
jlchang: Will this logged error message get passed to the user in the ingest notification email?

ingest/dense.py
```diff
@@ -74,6 +97,9 @@ def transform(self):
             self.info_logger.info(
                 f'Transforming gene :{gene}', extra=self.extra_log_params
             )
+            if gene in self.gene_names:
+                raise ValueError(f'Duplicate gene: {gene}')
```
jlchang: Will raising ValueError look like one of our scripts failed and exited?
Should this error be logged so the user receives information in the ingest notification email about why their parse failed?

ingest/dense.py
knapii-developments: All error log statements with get written to the same file and are passed to users.

ingest/dense.py
knapii-developments: The error 'Duplicate gene: {gene}' is passed to the log file and it presented in the email users get for parsing. Parsing fails when the duplicate gene is encountered.  I think that is the desired behavior.

PR #73, opened 2020-01-14
Test ingest for cluster files and metadata files (SCP-2026)

eweitz:
This adds tests for ingesting properly and improperly formatted cluster files and metadata files.  Test data was copied from [broadinstitute/single_cell_portal_core](http://github.com/broadinstitute/single_cell_portal_core).

These changes also:
* Fix a bug in error logging for metadata ingest
* Refactor `main()` in `ingest_pipeline.py` for better modularity and testability, and
* Enhance coverage of the entry and exit code for our CLI.

This satisfies SCP-2026.


eweitz:
> At some point we're going to need to test that the actual values we read out of files get transformed into the DataArray structures we expect them to be.

I concur, but it's worth noting that such tests of transformed model structure could be done at the interface or method levels.  Interface tests can be just as deep as method tests.

For example, we verify model structure in interface tests for expression ingest -- by intercepting calls to `load()` and evaluating what's passed into that method.

PR #74, opened 2020-01-15
Add technique for profiling memory (SCP-2053)

eweitz:
This enables developers to profile memory usage in Ingest Pipeline, via [memory_profiler](https://github.com/pythonprofilers/memory_profiler).

Out of memory errors cause scalability problems in Ingest Pipeline, and this tool gives us a way to identify which functions contribute to the problem (and how much).  For set-up instructions and preliminary results, see [Profiling memory in Ingest Pipeline](https://docs.google.com/document/d/1NobO9jDfX5NoAGxXoNdoqHoTZ83rTEJxL0IU_X91l-k/edit#).

Note: you must comment out all `@profile` decorators before pushing, or tests will fail.

This satisfies SCP-2053.


eweitz:
Regarding the new `cli_parser.py` module: this extracts miscellaneous functions that were in `ingest_pipeline.py`.  I refactored it while exploring a more ambitious implementation of memory profiling -- which would have enabled developers to profile remote runs, e.g. on PAPI.

I ended up reverting those changes as they're scoop creep, but separating those functions into a new module seems more maintainable anyways, so I retained that piece.

ingest/ingest_pipeline.py
```diff
@@ -202,7 +190,12 @@ def close_matrix(self):
         """Closes connection to file"""
         self.matrix.close()
 
+    # @profile
+    def insert_many(self, documents):
```
bistline: I think that we should probably also include the collection name for this, rather than hard-coding it to `data_arrays`.  For now, that is the only place we call `insert_many()`, but I have a feeling that it will change.

eweitz: Done in 5f53c3e.

eweitz:
> Also - it may be nice if we can have conditional decorators for profiling, rather than relying on commenting out code to prevent tests from failing. I found an example here: https://gist.github.com/T1T4N/22ca7b0764cefe917b2b3a6bb056364c

Agreed!  I've opened SCP-2081 to track that enhancement.

PR #75, opened 2020-01-22
Batch inserts for expression files (Implement batch inserts for MongoDB for genes in scp-ingest-pipeline)

knapii-developments:
Inserting documents has been identified as a performance bottleneck. To improve performance, ingest pipelines utilizes bulk writes to insert documents more quickly than before. A unit tests to ensure the transform function of Dense.py has been added to ensure the gene model is being created correctly. 

This PR satisfies [SCP-2063](https://broadworkbench.atlassian.net/browse/SCP-2063)


tests/test_ingest.py
```diff
@@ -123,20 +125,24 @@ def test_ingest_dense_matrix(self):
...
+        for model in models:
+            # Ensure that 'ObjectID' in model is removed
+            del model['_id']
+            self.assertEqual(model, gene_models[model['name']])
+        # print(models)
 
         # Verify gene model looks as expected
-        mock_dir = 'dense_matrix_19_genes_100k_cells_txt'
-        expected_model = get_gene_model(mock_dir)
+        # mock_dir = 'dense_matrix_19_genes_100k_cells_txt'
```
eweitz: Removing commented-out code here (and elsewhere in this file) would help readability.

ingest/ingest_pipeline.py
```diff
@@ -310,12 +311,38 @@ def upload_metadata_to_bq(self):
...
+            self.error_logger.error(e, extra=self.extra_log_params)
+            return False, None
+
     @trace
     @my_debug_logger()
     # @profile
     def ingest_expression(self) -> int:
         """Ingests expression files.
         """
+        gene_documents = []
```
bistline: It seems like this might be one place where we are experiencing a memory leak - I don't see where the contents of these lists get emptied.  It looks as though these two lists would just continue to grow as we read more data.

knapii-developments: @eweitz has already identified that this is a place for memory concerns in his [doc](https://docs.google.com/document/d/1NobO9jDfX5NoAGxXoNdoqHoTZ83rTEJxL0IU_X91l-k/edit). This ticket addresses speed and not memory. Further work will need to be done to address the memory leaks as identified in [SCP-1916](https://broadworkbench.atlassian.net/browse/SCP-1916)

bistline: OK, sounds good.

PR #78, opened 2020-01-31
capture metadata convention validation errors via logger (SCP-2092)

jlchang:
This modifies validation against metadata convention to use the logging functionality from monitor.py so metadata convention validation errors will be included in the errors.txt file allowing validation information to be passed to the end user in the parsing notification email.

Additionally, detection of discrepancy between the number of cells prepared for upload from the input metadata file and the number of rows inserted in the BigQuery job will now be reported in errors.txt.

This issue fulfills [SCP-2092](https://broadworkbench.atlassian.net/browse/SCP-2092).

### DIFF
--- a/ingest/validation/validate_metadata.py
+++ b/ingest/validation/validate_metadata.py
@@ -398,9 +398,10 @@ def collect_jsonschema_errors(metadata, convention, bq_json=None):
     # the latter two should be done together in the same pass thru the file
     js_errors = defaultdict(list)
     schema = validate_schema(convention, metadata)
-    # truncate jsonfile so appends invoked below are added to an empty file
+
     if bq_json:
         bq_filename = str(metadata.study_file_id) + '.json'
+        # truncate jsonfile so data from serialize_bq starts with an empty file
         fh = open(bq_filename, 'w')
         fh.close()
     if schema:
--- a/ingest/validation/validate_metadata.py
+++ b/ingest/validation/validate_metadata.py
@@ -426,14 +426,13 @@ def collect_jsonschema_errors(metadata, convention, bq_json=None):
         return False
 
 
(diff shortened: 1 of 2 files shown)


ingest/validation/validate_metadata.py
```diff
@@ -397,12 +393,16 @@ def collect_jsonschema_errors(metadata, convention, bq_json=None):
...
     returns False if input convention is invalid JSON schema
     """
-    logger.debug('Begin: collect_jsonschema_errors')
     # this function seems overloaded with its three tasks
     # schema validation, non-ontology errors, ontology info collection
     # the latter two should be done together in the same pass thru the file
     js_errors = defaultdict(list)
     schema = validate_schema(convention, metadata)
+    # truncate jsonfile so appends invoked below are added to an empty file
+    if bq_json:
```
devonbush: I'd move the # truncate comment inside the 'if' block

jlchang: refined comment to indicate method of subsequent appends for clarity
288698f

ingest/validation/validate_metadata.py
```diff
@@ -397,12 +393,17 @@ def collect_jsonschema_errors(metadata, convention, bq_json=None):
...
-    logger.debug('Begin: collect_jsonschema_errors')
     # this function seems overloaded with its three tasks
     # schema validation, non-ontology errors, ontology info collection
     # the latter two should be done together in the same pass thru the file
     js_errors = defaultdict(list)
     schema = validate_schema(convention, metadata)
+
+    if bq_json:
+        bq_filename = str(metadata.study_file_id) + '.json'
+        # truncate jsonfile so data from serialize_bq starts with an empty file
```
```suggestion
        # truncate JSON file so data from serialize_bq starts with an empty file
```
```suggestion
        # truncate JSON file so data from serialize_bq starts with an empty file
```

PR #79, opened 2020-02-12
Regularize stored ontology IDs (SCP-2116)

jlchang:
Refactored data preparation for BigQuery upload to consistently store ontology IDs using the underscore format to simplify metadata query on ontology IDs.

Additional changes include:
- additional error messaging where ingest pipeline load fails due to lack of MongoDB access
- fix lurking issue where single value arrays were not being cast to convention-specified data type
- fix lurking issue where cast_integer_type would fail if value was nan
- avoid casting nan to string because nan is indicative of missing values
- update reference json for metadata validation tests due to changes in nan handling

NOTE: current JSON Schema validation implementation REQUIRES that non-float convention metadata cannot have nan values. SCP-2137 has been created to address this unintended requirement which becomes blocking for metadata files from cell-based advanced search results.

This PR satisfies [SCP-2116](https://broadworkbench.atlassian.net/browse/SCP-2116)

### DIFF
--- a/ingest/ingest_pipeline.py
+++ b/ingest/ingest_pipeline.py
@@ -229,7 +229,7 @@ class IngestPipeline(object):
             self.error_logger.error(e, extra=self.extra_log_params)
             try:
                 self.error_logger.error(e.details, extra=self.extra_log_params)
-            except AttributeError as e:
+            except AttributeError:
                 self.error_logger.error(
                     'Check access to MongoDB', extra=self.extra_log_params
                 )
--- a/ingest/ingest_pipeline.py
+++ b/ingest/ingest_pipeline.py
@@ -227,13 +227,12 @@ class IngestPipeline(object):
                 self.insert_many('data_arrays', documents)
         except Exception as e:
             self.error_logger.error(e, extra=self.extra_log_params)
-            self.error_logger.error(
-                'Unable to access to MongoDB?', extra=self.extra_log_params
-            )
             try:
                 self.error_logger.error(e.details, extra=self.extra_log_params)
             except AttributeError as e:
-                self.error_logger.error(e, extra=self.extra_log_params)
+                self.error_logger.error(
+                'Check access to MongoDB', extra=self.extra_log_params
+            )
             return 1
         return 0
 
--- a/ingest/ingest_pipeline.py
+++ b/ingest/ingest_pipeline.py
@@ -231,8 +231,8 @@ class IngestPipeline(object):
                 self.error_logger.error(e.details, extra=self.extra_log_params)
             except AttributeError as e:
                 self.error_logger.error(
-                'Check access to MongoDB', extra=self.extra_log_params
-            )
+                    'Check access to MongoDB', extra=self.extra_log_params
+                )
             return 1
         return 0
 
--- a/ingest/ingest_pipeline.py
+++ b/ingest/ingest_pipeline.py
@@ -227,8 +227,11 @@ class IngestPipeline(object):
                 self.insert_many('data_arrays', documents)
         except Exception as e:
             self.error_logger.error(e, extra=self.extra_log_params)
-            if e.details is not None:
+            self.error_logger.error('Unable to access to MongoDB?', extra=self.extra_log_params)
+            try:
                 self.error_logger.error(e.details, extra=self.extra_log_params)
+            except AttributeError as e:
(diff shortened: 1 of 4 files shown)


ingest/validation/validate_metadata.py
```diff
@@ -332,6 +364,12 @@ def cast_metadata_type(metadatum, value, id_for_error_detail, convention, metada
             # files that support array-based metadata navtively (eg. loom,
             # anndata etc) splitting on pipe may become problematic
             for element in value.split('|'):
+                try:
+                    if 'ontology' in convention['properties'][metadatum]:
+                        element = regularize_ontologyID(element)
+                except KeyError:
+                    # non-metadata convention metadata will trigger this exception
```
eweitz: "Non-metadata-convention" is somewhat wordy.  How about this?

```suggestion
                    # unconventional metadata will trigger this exception
```

While "unconventional" by itself is unclear, but "unconventional metadata" seems fairly unambiguous.

If you agree, a global find/replace would make this change consistently.

ingest/validation/validate_metadata.py
```diff
@@ -307,12 +322,29 @@ def cast_float_type(value):
...
+        # nan indicates missing data; by type, nan is a numpy float
+        # so a separate type check is needed for proper handling
+        return value
+    elif isinstance(value, numbers.Number):
         return str(value)
     else:
         return value
 
 
+def regularize_ontologyID(value):
```
eweitz: Simpler, more Python-conventional casing:

```suggestion
def regularize_ontology_id(value):
```

A global find/replace from "ontologyID" to "ontology_id" would make this change consistently.

ingest/validation/validate_metadata.py
```diff
@@ -307,12 +322,29 @@ def cast_float_type(value):
...
+    """Regularize ontologyIDs for storage with underscore format
+    """
+    try:
+        ontology_shortname, term_id = re.split('[_:]', value)
+        value = ontology_shortname + '_' + term_id
+        return value
+    # when ontolgyID is malformed and has no separator -> ValueError
```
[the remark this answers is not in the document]
jlchang: addressed in 639a006

ingest/validation/validate_metadata.py
jlchang: addressed in 639a006

ingest/validation/validate_metadata.py
jlchang: addressed in 639a006

ingest/validation/validate_metadata.py
```diff
@@ -307,12 +322,29 @@ def cast_float_type(value):
...
         return str(value)
     else:
         return value
 
 
+def regularize_ontology_id(value):
+    """Regularize ontology_ids for storage with underscore format
+    """
+    try:
+        ontology_shortname, term_id = re.split('[_:]', value)
```
devonbush: I think it would be easier instead of splitting and rejoining to just do
`return value.replace(":", "_")`

ingest/validation/validate_metadata.py
```diff
@@ -307,12 +322,29 @@ def cast_float_type(value):
...
+        value = ontology_shortname + '_' + term_id
+        return value
+    # when ontology_id is malformed and has no separator -> ValueError
+    # when ontology_id value is empty string -> TypeError
+    except (ValueError, TypeError):
+        return value
```
devonbush: wouldn't we want to fail here, instead of returning the value?

ingest/validation/validate_metadata.py
jlchang: great suggested, implemented in 3ad4b5d

ingest/ingest_pipeline.py
```diff
@@ -227,8 +227,12 @@ def load(
                 self.insert_many('data_arrays', documents)
         except Exception as e:
             self.error_logger.error(e, extra=self.extra_log_params)
-            if e.details is not None:
```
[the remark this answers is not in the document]
jlchang: refactored in 3ad4b5d

ingest/validation/validate_metadata.py
jlchang: added comment in 3ad4b5d to explain that we return the bad value so it can be evaluated with the JSON Schema validator so the end user receives an actionable error message within the context of metadata validation

PR #82, opened 2020-03-31
Accommodate additional valid ontology IDs and update metadata convention location (SCP-2230)

jlchang:
This work allows ontologyID syntax to use the less strict format ^[-A-Za-z0-9]+[_:][-A-Za-z0-9]+
allowing valid ontologyIDs (eg NCIT IDs) to pass validation.

Additionally this work formalizes a repository location for metadata convention under schema/<project>_convention

serialize_convention.py has been updated to leverage the formal convention location in the repo and reflects the convention version in the convention json (ie. the $id field) so version information will be available to be put into BigQuery.

This fulfills SCP-2230


scripts/serialize_convention.py
```diff
@@ -151,23 +141,21 @@ def write_json_schema(filename, object):
         jsonfile.write(object)
 
 
-def generate_output_name(inputname, label, path=''):
+def check_if_old_output(project, version):
```
devonbush, line 144: since this doesn't just return a true/false check, consider renaming to get_input_output_names or something similar

jlchang: addressed in `41272306`

PR #85, opened 2020-04-13
Fix metadata of type "group" behaving as float post-ingest

knapii-developments:
Prior, ingest service did not account for numeric like group labels  in cluster and metadata files. As a result, numeric group labels were treated as floats/integers instead of strings. This PR fixes this resolves the issue as well as tests to ensure column labels are treated as strings. 

To test please download the [PigCluster file](https://single-cell-staging.broadinstitute.org/single_cell/data/private/SCP50/zen4840?filename=Pig_clusterfile.txt) and run  
``` 
python ingest_pipeline.py --study-id 5d276a50421aa9117c982845 --study-
file-id 5dd5ae25421aa910a723a337 ingest_cell_metadata --cell-metadata-file <path to file> --stud
y-accession SCP123 --ingest-cell-metadata 
```

This PR satisfies [SCP-2214](https://single-cell-staging.broadinstitute.org/single_cell/data/private/SCP50/zen4840?filename=Pig_metafile.csv)


ingest/ingest_files.py
```diff
@@ -270,30 +272,11 @@ def open_txt(self, open_file_object, **kwargs):
     def open_pandas(self, file_path, file_type, **kwargs):
         """Opens file as a dataframe """
         open_file_object = kwargs.pop('open_file_object')
-        if file_type == "text/tab-separated-values":
```
jlchang, line 273: Worried that changing how we open tsv and csv files will have subtle changes to the behavior of ingest pipeline that our tests won't detect. We should get this change into staging (especially if Sarah is willing to upload some 2.0.0 data into staging) and consider uploading a variety of known good files (csv/txt and also rename to txt) and make sure we don't think this change has unintended side effects.

knapii-developments:
> Looks good, but a concern with the manual test:
> 
> I recall from pairing yesterday that this issue was seen in a .csv file (`Pig_metafile.csv`) and not a .txt file (`Pig_clusterfile.txt`) as currently linked in PR description. I could be mis-remembering, but I think `Pig_metafile.csv` is the appropriate test data file.
> 
> To ensure the manual test instructions don't give a false negative (i.e. passing test if this fix were to regress), could you verify that the bug affected `Pig_clusterfile.txt`? If it didn't, the instructions should be updated to use `Pig_metafile.csv`.
> 
> Also it'd be nice to pare down the test data file to a smaller number of rows that still manifest this issue, and add that to this repo -- links to staging shouldn't be considered stable. This would also ease writing an automated test for this (though I don't consider that blocking).

Your assumption is correct. I linked the incorrect file. I've updated it now

PR #89, opened 2020-04-29
Handle nan __ontology_label values (SCP-2375)

jlchang:
Handle edge case where metadata file provides an __ontology_label column but does not provide values (prevent validation from making excessive EBI OLS calls due to bug by not catching this edge case).

For required metadata, JSON schema validation will fail because 'nan' is not a string - apparently supplying nan does not count as violating the "required" constraint. Note that for optional metadata, nan values are not passed in JSON Schema for validation, only a cross-check warning will be emitted.

This commit includes updates to a reference test output file so reference matches new parse behavior.

This satisfies SCP-2375

### DIFF
--- a/ingest/validation/validate_metadata.py
+++ b/ingest/validation/validate_metadata.py
@@ -203,30 +203,48 @@ def validate_cells_unique(metadata):
     return valid
 
 
-def collect_cell_for_ontology(metadatum, row_data, metadata, array=False):
+def collect_cell_for_ontology(metadatum, row, metadata, array=False):
     """Collect ontology info for a single metadatum into CellMetadata.ontology dictionary
     """
     if metadatum.endswith('__unit'):
         ontology_label = metadatum + '_label'
     else:
         ontology_label = metadatum + '__ontology_label'
+    cell_id = row['CellID']
     if array:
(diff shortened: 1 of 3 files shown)


ingest/validation/validate_metadata.py
```diff
@@ -203,30 +203,48 @@ def validate_cells_unique(metadata):
...
+                metadata.ontology[metadatum][(id)].append(cell_id)
         except (TypeError, KeyError):
-            for id in row_data[metadatum]:
-                metadata.ontology[metadatum][(id)].append(row_data['CellID'])
+            # Not all ontology data will have user-provided ontology_label
+            # where provided, collect to provide cross-validation
+            try:
+                ontology_dict = dict(zip(row[metadatum], row[ontology_label]))
+                for id, label in ontology_dict.items():
+                    metadata.ontology[metadatum][(id, label)].append(cell_id)
```
eweitz, line 227: Renaming `metadata.ontology[metadatum]` to something shorter would reduce visual noise, which would maintainability for this complex method.

You could append to that more ergonomic variable instead of `metadata.ontology[metadatum]`, then `return` it at the end of the method.

PR #90, opened 2020-04-29
Update alexandria metadata convention to 2.0.1 (SCP-2376)

jlchang:
Two metadata convention errors detected during testing
1. enum for preservation method had typo "Snap", "Frozen" instead of "Snap Frozen"
2. extended ontologyID syntax only applied to arrays and not single objects

This work updates the metadata convention and associated tests to satisfy SCP-2376


ingest/ingest_pipeline.py
```diff
@@ -92,7 +92,7 @@ class IngestPipeline(object):
     JSON_CONVENTION = (
         '../schema/alexandria_convention/alexandria_convention_schema.json'
     )
-    EXTERNAL_JSON_CONVENTION = 'gs://broad-singlecellportal-public/schema/alexandria_convention/2.0.0/alexandria_convention_schema.json'
+    EXTERNAL_JSON_CONVENTION = 'gs://broad-singlecellportal-public/schema/alexandria_convention/2.0.1/alexandria_convention_schema.json'
```
bistline, line 95: We're keeping this until we update `manage_study.py` to use the SCP REST API, correct?

jlchang: Yes, I checked with Eno and we agreed that it wouldn't make sense for me to remove `EXTERNAL_JSON_CONVENTION` until `manage_study.py` was using the API endpoint to obtain the metadata convention.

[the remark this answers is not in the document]
jlchang: I checked with Eno and we'll remove `EXTERNAL_JSON_CONVENTION` as soon as manage_study.py is no longer using a google bucket.

PR #94, opened 2020-05-12
Validate against Mouse Brain Atlas ontology file (SCP-2361)

jlchang:
- Validation of organ_region using ontology file MouseBrainAtlas.csv stored in this repo. 
- Updated metadata convention to 2.1.1 to improve readability of cell_type__custom values.
- Automated test to show validation being successfully applied on invalid organ_region and organ_region__ontology_label inputs.

This satisfies SCP-2361


ingest/validation/validate_metadata.py
```diff
@@ -808,6 +822,60 @@ def validate_collected_ontology_data(metadata, convention):
     return
 
 
+def validate_organ_region_metadata(ontology_info, region_ontology, metadata):
+    """
+    Validation for nonEBI-OLS ontology, Mouse Brain Atlas
+    """
```
eweitz, line 828: More [PEP8-conventional](https://www.python.org/dev/peps/pep-0008/#documentation-strings):

```suggestion
    """Validation for non-EBI-OLS ontology, Mouse Brain Atlas
    """
```

And similarly elsewhere.

ingest/validation/validate_metadata.py
```diff
@@ -808,6 +822,60 @@ def validate_collected_ontology_data(metadata, convention):
...
+                metadata.store_validation_issue('error', 'ontology', error_msg)
+
+
+def parse_organ_region_ontology_id(ontology_id, metadata):
+    """
+    Check identifier
+    """
+    try:
+        ontology_shortname, term_id = re.split('[_:]', ontology_id)
+        if ontology_shortname == "MBA":
```
eweitz, line 865:
```suggestion
        if ontology_shortname == 'MBA':
```

ingest/validation/validate_metadata.py
```diff
@@ -726,10 +727,23 @@ def validate_collected_ontology_data(metadata, convention):
     for entry in metadata.ontology.keys():
         # provision to skip validation of non-EBI OLS metadata for now
         if entry == 'organ_region':
-            continue
+            MBA_url = 'https://raw.githubusercontent.com/broadinstitute/scp-ingest-pipeline/jlc_validate_organ_region/schema/organ_region/mouse_brain_atlas/MouseBrainAtlas.csv'
```
eweitz, line 730: Blocker: There's an interesting predicament here.  

**Problem 1**
If someone deletes the innocuously-named `jlc_validate_organ_region` branch, then to my undertanding the current `MBA_url` value will return a 404 and production will break.

One solution would be to update `MBA_url` to point to the `master` branch.

**Problem 2**
Updating to `master` means that your test using this ontology would fail before this branch is merged.

---

So both approaches seem problematic, in different ways, and one problem can't be resolved without causing the other.  We can avoid both by pointing to a particular commit ID, but then we sacrifice the ability to always point to the latest version.

Solving Problem 1 instead of (the merely ceremonial) Problem 2 strikes me as a better trade-off.  For example:

```suggestion
            MBA_url = 'https://raw.githubusercontent.com/broadinstitute/scp-ingest-pipeline/master/schema/organ_region/mouse_brain_atlas/MouseBrainAtlas.csv'
```

ingest/validation/validate_metadata.py
```diff
@@ -726,10 +727,23 @@ def validate_collected_ontology_data(metadata, convention):
     for entry in metadata.ontology.keys():
         # provision to skip validation of non-EBI OLS metadata for now
```
bistline, line 728: This comment isn't true anymore

ingest/validation/validate_metadata.py
bistline, line 730: I think that we probably want a better place to store this - loading from a GitHub URL on a branch will depend on that branch always existing.  Also, if this is in the same repository, why are you loading it remotely, and not from its local path?

jlchang: Thanks for catching this issue, Eric. My current thinking is that this is ontology data and should only change rarely.  It is less important that we point to latest and more important that the code is always pointed to the ontology file that will give correct validation. In that sense, the convention in the repo at this point in time (ie. for commit X) can be tied to an appropriate instance of the ontology file for a pre-existing commit.

jlchang: validate_collected_ontology_data gets called from both test_validate_metadata.py and from manage_study.py. When called in those places, relative paths don't work.

I'm planning to update the URL to point to a particular commit which will slightly address the URL issue for now. Devon is helping me look into how we might use the local copy and not need to load the file from GitHub. We would like to make this more robust and not depend on an external source but didn't want that refinement to block getting the basic functionality into place.

ingest/validation/validate_metadata.py
```diff
@@ -726,10 +727,23 @@ def validate_collected_ontology_data(metadata, convention):
...
         if entry == 'organ_region':
-            continue
+            MBA_url = 'https://raw.githubusercontent.com/broadinstitute/scp-ingest-pipeline/jlc_validate_organ_region/schema/organ_region/mouse_brain_atlas/MouseBrainAtlas.csv'
+            try:
+                region_ontology = pd.read_csv(MBA_url, header=0)
+            # if we can't get the file in GitHub for validation, exit
+            except:  # noqa E722
+                error_msg = f'Unable to read GitHub-hosted organ_region ontology file'
```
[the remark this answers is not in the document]
jlchang: `bf48ed8`

ingest/validation/validate_metadata.py
jlchang: `bf48ed8`

ingest/validation/validate_metadata.py
jlchang: `bf48ed8`

ingest/validation/validate_metadata.py
jlchang: `bf48ed8` updated to point to first commit where file was added to repo

ingest/validation/validate_metadata.py
jlchang: `bf48ed8` updated

ingest/validation/validate_metadata.py
jlchang: `bf48ed8` updated to point to first commit where file was added to repo

PR #101, opened 2020-06-03
Ingest cell metadata returns non-integer status and fails

knapii-developments:
This PR addresses a bug reported from [zendesk](https://broadinstitute.zendesk.com/agent/tickets/5748). The user had NA values in a column that should've been treated as an object (strings are considered objects). Pandas infers NA as NaN (not a number) and interprets the value as a float. As a result, subsampling failed when merging a cluster and metadata file because you can not perform a merge on different dtypes (in this case merging an object column on a float). Dtypes are now assigned to columns based on annotation type. 

Secondly, emails were not being sent when subsampling failed because exceptions were not caught properly. Exceptions are now caught for subsampling and the exit code is successfully returned. A bug in test_annotations was fixed to ensure the merge function is working properly.

This ticket satisfies [SCP-2454](https://broadworkbench.atlassian.net/browse/SCP-2454)


ingest/annotations.py
```diff
@@ -32,13 +32,12 @@ def __init__(
...
-            file_path, open_as='dataframe', header=[0, 1]
-        )
-        self.headers = self.file.columns.get_level_values(0)
-        self.annot_types = self.file.columns.get_level_values(1)
         # lambda below initializes new key with nested dictionary as value and avoids KeyError
         self.issues = defaultdict(lambda: defaultdict(lambda: defaultdict(list)))
+        csv_file, self.file_handle = self.open_file(self.file_path)
+        # Remove white spaces, quotes, and lowercase
+        self.headers = [header.strip().strip('\"').lower() for header in next(csv_file)]
+        self.annot_types = [type.strip().strip('\"').lower() for type in next(csv_file)]
```
jlchang, line 40: I'm happy to see us ignore accidental whitespace in the SCP headers, it avoids a failure mode  that is hard to detect by eye.

ingest/annotations.py
```diff
@@ -73,38 +72,40 @@ def merge_df(self, first_df, second_df):
...
+            - Labels are treated as strings (objects)
+            - Numeric annotations are treated as float32
+            - Numeric columns are rounded to 3 decimals points
+        """
+        columns = []
+        dtypes = {}
+        dtypes[self.headers[0]] = 'object'
+        dtypes[self.annot_types[0]] = 'object'
+        for annotation, annot_type in zip(self.headers, self.annot_types):
+            dtypes[annotation] = 'object' if annotation is 'group' else 'float32'
```
eweitz, line 96: I've had floats on my mind likely, so I wonder: would [`float16`](https://pandas.pydata.org/pandas-docs/stable/user_guide/gotchas.html#why-not-make-numpy-like-r) suffice here?  If not, are we sure `float64` isn't needed?

Using smaller datatypes until we know we need bigger ones seems like would help constrain memory.

While memory usage is an issue with expression ingest, it's not a known problem with metadata ingest, so this comment is merely something to consider -- definitely not a blocker.

ingest/validation/validate_metadata.py
```diff
@@ -948,6 +948,7 @@ def validate_input_metadata(metadata, convention, bq_json=None):
     """Wrapper function to run validation functions
     """
     collect_jsonschema_errors(metadata, convention, bq_json)
+    print(metadata.issues)
```
jlchang, line 951: This looks like it was added for debugging? Please remove before merge.

tests/test_annotations.py
```diff
@@ -72,14 +72,17 @@ def test_merge_df(self):
         )
         cell_metadata_df.preprocess()
         cell_names_cell_metadata_df = np.asarray(cell_metadata_df.file['NAME'])
+        print(cell_names_cell_metadata_df)
```
jlchang, line 75: This print statement (and the next three print statements) look like they may be for debugging?

ingest/annotations.py
knapii-developments: I originally put float32. I think it might. However, I'm more worried about precision

ingest/annotations.py
```diff
@@ -73,38 +72,40 @@ def merge_df(self, first_df, second_df):
...
-            ).columns.tolist()
-            self.file[group_columns] = self.file[group_columns].astype(str)
-        # Find numeric columns and round to 3 decimals places and are floats
-        # only run this assignment if numeric annotations are present
-        if 'numeric' in list(annot_types):
+        self.headers[0] = self.headers[0].upper()
+        self.annot_types[0] = self.annot_types[0].upper()
+        self.createDataFrame()
+
+    def createDataFrame(self):
```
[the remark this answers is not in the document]
knapii-developments: 553dd81

ingest/validation/validate_metadata.py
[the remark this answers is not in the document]
knapii-developments: It was left over. 553dd81

knapii-developments: 553dd81

tests/test_annotations.py
knapii-developments: 553dd81

PR #102, opened 2020-06-08
Move /scripts/genomes from single_cell_portal to scp-ingest-pipeline (SCP-2470)

eweitz:
This copies code [from `single_cell_portal`](https://github.com/broadinstitute/single_cell_portal/tree/0646433c10c34cf91de531de82f1f6861b3bb8de/scripts/) to Ingest Pipeline.  Code for generating species and genome reference data has been moved from `/scripts/genomes` there to a new directory here, `/ingest/genomes`.  And `make_toy_data.py` has been copied to `/ingest` here.

The genome pipeline in `/ingest/genomes` here will need to be updated to fix an igv.js issue.  That pipeline isn't intended for technical end-users, so it doesn't fit in `single_cell_portal`, and having that and `make_toy_data.py` here simplifies dependency management for `make_toy_data.py`.  More background is available in broadinstitute/single_cell_portal#99.

To complete the move, I've also opened a [PR in `single_cell_portal_core`](broadinstitute/single_cell_portal_core#621) and a [PR in `single_cell_portal`](broadinstitute/single_cell_portal#101) to remove the slightly different(!) copies of this code in those places.  That will resolve the biggest tech debt in the genome pipeline and satisfy SCP-2470.  Other pre-existing tech debt is documented and scoped in SCP-2491.

This PR and the two noted above satisfy SCP-2470.


ingest/genomes/parse_genome_annotations.py
```diff
@@ -0,0 +1,250 @@
...
+
+    return [gtf_urls, ensembl_metadata]
+
+
+def transform_ensembl_gtf(gtf_path, ref_dir):
+    """Produce sorted GTF and GTF index from Ensembl GTF; needed for igv.js
+    """
+    # Example:
+    # $ sort -k1,1 -k4,4n gencode.vM17.annotation.gtf > gencode.vM17.annotation.possorted.gtf
+    # $ bgzip gencode.vM17.annotation.possorted.gtf
```
bistline, line 82: I have the same comment here that I had in https://github.com/broadinstitute/single_cell_portal/pull/99/files#r434173376.  You mentioned a README update in this repo to address it?

eweitz: Done in 11ecf0a.  I had scoped that update among other pre-existing tech debt in SCP-2491, but this was a trivial change.

PR #103, opened 2020-06-09
Publish scp-ingest-pipeline to PyPI (SCP-2471)

jlchang:
Minor updates to publish scp-ingest-pipeline to PyPI

These changes support SCP-2471.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,8 +4,8 @@ with open("README.md", "r") as fh:
     long_description = fh.read()
 
 setup(
-    name='scp-ingest-pipe',
-    version='1.3.6rc1',
+    name='scp-ingest-pipeline',
+    version='1.3.7rc1',
     description='ETL pipeline for single-cell RNA-seq data',
     long_description=long_description,
     long_description_content_type="text/markdown",
--- a/setup.py
+++ b/setup.py
@@ -5,7 +5,7 @@ with open('README.md', 'r') as fh:
 
 setup(
     name='scp-ingest-pipeline',
-    version='1.3.7rc1',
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -1,8 +1,14 @@
 from setuptools import setup, find_packages
+
+with open("README.md", "r") as fh:
+    long_description = fh.read()
+
 setup(
-    name='scp_ingest_pipeline',
-    version='0.1',
+    name='scp-ingest-pipe',
+    version='1.3.6rc1',
```
jlchang, line 8: The repo is currently on release 1.3.6
The state of master has code that will be in release 1.3.7 - should our version be 1.3.6rc1 or 1.3.7rc1?

eweitz: `1.3.7rc1`, I think.

Also, `scp-ingest-pipeline` instead of `scp-ingest-pipe`.

setup.py
```diff
@@ -24,8 +30,15 @@
...
         'opencensus-ext-stackdriver',
         'google-cloud-trace',
-	'grpcio',
-	'grpcio-tools'
+        'grpcio',
+        'grpcio-tools',
     ],
     packages=find_packages(),
+    classifiers=[
+        'Development Status :: 5 - Production/Stable',
```
devonbush, line 38: nice, I think this is right -- not beta, but also not 'mature' :)

setup.py
jlchang: Thanks for catching scp-ingest-pipe! Fixed and using 1.3.7rc1 in `9a25885`

PR #106, opened 2020-06-15
Ensure .txt delimiter detection and group annotation behavior (SCP-2498, SCP-2508)

jlchang:
Recent changes in ingest pipeline caused two issues:
1.  tab delimited .txt files are no longer ingested and error with "Could not determine delimiter" message
2. group annotations whose values are detected as numeric by Pandas cannot load to MongoDB and error with "cannot encode object: 1, of type: "

Code change to explicitly assign dtypes to dataframe subtly changed ingest file handling.

Previously, all annotation files (including .txt) were opened as dataframe by Pandas where delimiters were detected based on the first line of the file. Current code now uses open_txt for .txt files. Modification of open_txt to also detect delimiters with the first line rather than the first 1024 bytes of the file should restore delimiter detection. Changed suffix of input files for several existing tests from .tsv to .txt to ensure regressions will be detected.

Previously, annotation dtypes were inferred by Pandas and group annotation behaviour was applied to Pandas-inferred numeric values after the preprocessing step. The assumption that application of "object" dtype contents of a Pandas dataframe to become strings is false. (for detailed example, refer to https://pbpython.com/currency-cleanup.html). The solution applied is to explicitly coerce columns within a Pandas dataframe to type string AND make sure that reset_file does not overwrite the explicit coercion. Testing for coercion has been updated to test on the dataframe column's dtype and use np.issubdtype.

This PR supports bugfixes for SCP-2498 and SCP-2508.


ingest/annotations.py
```diff
@@ -98,6 +96,14 @@ def create_data_frame(self):
         self.file = self.open_file(
             self.file_path, open_as='dataframe', dtype=dtypes, names=index, skiprows=2
         )[0]
+        # dtype of object allows mixed dtypes in columns, including numeric dtypes
+        # coerce group annotations pandas detects as non-object types to type string
```
eweitz, line 100: I had to re-read this line a few times to understand.  A tiny tweak would help readability.

```suggestion
        # coerce group annotations that pandas detects as non-object types to type string
```

ingest/ingest_files.py
```diff
@@ -278,12 +278,19 @@ def open_txt(self, open_file_object, file_type, **kwargs):
...
+        try:
+            csv_dialect = csv.Sniffer().sniff(
+                open_file_object.readline(), delimiters=delimiter
+            )
+            csv_dialect.skipinitialspace = True
+            open_file_object.seek(0)
+            return csv.reader(open_file_object, csv_dialect)
+        except Exception:
+            raise ValueError(
+                f'Could not determine delimiter. Please save file with appropriate suffix (.tsv or .csv) and try again.'
```
bistline, line 292: I like this enhancement - especially if it can be surfaced and returned to the user as the main error message.

PR #26, opened 2019-09-18
Integrate Firestore emulator

eweitz:
This integrates the [Firestore emulator](https://cloud.google.com/sdk/gcloud/reference/beta/emulators/firestore/) into automated tests for Ingest Pipeline.  Benefits include:

* Faster development.  Rather than writing to a remote Firestore, engineers can write to a local Firestore.  Manual tests indicate a roughly 4x speedup (~30 seconds before, ~7 seconds after) for our [test matrix](https://github.com/broadinstitute/scp-ingest-pipeline/commit/bbfa5d4d3e6e0afa6d248201eeecee8d40238758#diff-b77fbe45bbe85acb9b32cc7a810d777b).

* More code coverage.  Before, we could not quickly test the "load" phase of this ETL pipeline.  Now we do.  Compare coverage [before](https://249-187902423-gh.circle-artifacts.com/0/home/circleci/scp-ingest-pipeline/tests/htmlcov/_home_circleci_scp-ingest-pipeline_ingest_ingest_pipeline_py.html#n108) and [after](https://258-187902423-gh.circle-artifacts.com/0/home/circleci/scp-ingest-pipeline/tests/htmlcov/_home_circleci_scp-ingest-pipeline_ingest_ingest_pipeline_py.html#n112).  This enables fast, reliable integration tests.

To set up Firestore emulator locally, ensure you have Java (JRE >=8) installed, then:
* `gcloud components install beta cloud-firestore-emulator`
* `gcloud beta emulators firestore start --host-port localhost:8081`
* In another Terminal:
`export FIRESTORE_EMULATOR_HOST=localhost:8081`
* Run your development command, e.g.:
`python3 ingest_pipeline.py  --study-accession SCP1 --file-id 123abc ingest_expression --taxon-name 'Homo sapiens' --taxon-common-name human --ncbi-taxid 9606 --matrix-file ../tests/data/dense_matrix_19_genes_100k_cells.txt --matrix-file-type dense`

That approach works for Ingest Pipeline development.

Development on the Rails codebase could theoretically also then point to this local Firestore, assuming the file metadata being kept in MongoDB were already populated.  The approach described in the [`google-cloud-ruby`](https://github.com/googleapis/google-cloud-ruby/blob/master/google-cloud-firestore/EMULATOR.md#google-cloud-firestore-emulator) library is probably the best starting point, if we pursued that.  The steps above for "set up Firestore locally" could likely be used in place of the Firebase-specific instructions in that `google-cloud-ruby` link.

[...]

PR #108, opened 2020-06-17
Ingest 200k dense matrix

knapii-developments:
The PR is a 2 part series for ingesting a 200k dense matrix. The PR introduces a major _design change_ that complements the way ingest pipeline _behaves_. **Given the approval of the PR, the tests will be written and merged prior to merging this PR into master.**

**A command design pattern has been implemented to**:

- Specify, queue, and execute particular methods in a given order and or time.
- Parameterize objects by an action to perform. This form of action encapsulation allows additional functions related to the executed actions to be added
- Make classes reusable when it interacts with other classes. This makes things easier to test. This is done using [Dependency Injection](https://medium.com/@shivama205/dependency-injection-python-cb2b5f336dce).

**!!!** I accidentally didn't perform a git move for dense.py. Only part that's changed is the [transform function](https://github.com/broadinstitute/scp-ingest-pipeline/pull/108/files#diff-f334073370f494f6dbaa9d6db4307193R100) and added variables in[ __init__](https://github.com/broadinstitute/scp-ingest-pipeline/pull/108/files#diff-f334073370f494f6dbaa9d6db4307193R37) **!!!**


ingest/expression_files/dense_command.py
```diff
@@ -0,0 +1,30 @@
+import abc
+import sys
+
+from command import Command
+from expression_files.dense import Dense
+
+sys.path.append("../ingest")
```
eweitz, line 7: Typically this is put above certain imports, but given no imports are below it this line seems removable.

```suggestion
```

ingest/expression_files/expression_files.py
```diff
@@ -85,15 +94,17 @@ def set_data_array_gene_cell_names(
             linear_data_id,
             self.study_id,
             self.study_file_id,
-        ).get_data_array()
+        ).get_data_array():
+            yield model
```
eweitz, line 98: `set_data_array_cells` and `set_data_array_gene_cell_names` looks mostly redundant.  

Using `self.cluster_name` if `name` is `None` or some such seems like a possible way to consolidate these two methods.

If consolidation is unworkable, then using a consistent order for the two methods' seems worthwhile.  (One has `values` then `linear_data_id` in its signature, the other has `linear_data_id` then `values`; using a single consistent ordering seems more maintainable.)

ingest/expression_files/expression_files.py
```diff
@@ -102,4 +113,46 @@ def set_data_array_gene_expression_values(
...
+            raise Exception(f'{e}')
+
+        # Try writing gene docs
+        try:
+            gene_doc_bulk_write_results = self.mongo_connection.client[self.COLLECTION_NAME].bulk_write(
+                gene_model_bulk_operations,  ordered=False
+            )
+        except BulkWriteError as bwe:
+            print(f"error caused by gene docs : {bwe.details}")
+            raise BulkWriteError(f'Error caused by data docs : {bwe.details}')
```
eweitz, line 153: These seem like they should agree:

```suggestion
            print(f"error caused by gene docs : {bwe.details}")
            raise BulkWriteError(f'Error caused by gene docs : {bwe.details}')
```

ingest/ingest_pipeline.py
```diff
@@ -33,42 +33,43 @@
...
-from pymongo.errors import BulkWriteError
-
+from cell_metadata import CellMetadata
+from cli_parser import create_parser, validate_arguments
+from clusters import Clusters
+from expression_files.dense import Dense
+from expression_files.dense_command import DenseCommand
 # from google.cloud.logging.resource import Resource
-
+#
```
eweitz, line 54:
```suggestion

```

ingest/ingest_pipeline.py
```diff
@@ -153,7 +155,8 @@ def get_mongo_db(self):
         user = os.environ['MONGODB_USERNAME']
         password = os.environ['MONGODB_PASSWORD']
         db_name = os.environ['DATABASE_NAME']
```
eweitz, line 157: Will `MongoConnection()` replace `def get_mongo_db()`?  I assume so.

Could you add a comment in the latter if it'll be replaced, or if not what makes its use case different from `MongoConnection()`?

ingest/ingest_pipeline.py
```diff
@@ -546,8 +479,8 @@ def exit_pipeline(ingest, status, status_cell_metadata, arguments):
                             'errors.txt',
                             f'parse_logs/{study_file_id}/errors.txt',
                         )
-                    # Need 1 argument that has a path to identify google bucket
-                    # Break after first argument
+                    Need 1 argument that has a path to identify google bucket
+                    Break after first argument
```
eweitz, line 483: Blocker: leaving these lines uncommented looks like a functional bug.  Seemingly trivial fix:

```suggestion
                    # Need 1 argument that has a path to identify google bucket
                    # Break after first argument
```

ingest/invoker.py
```diff
@@ -0,0 +1,31 @@
+from command import Command
+
+
```
devonbush, line 3: This class seems unnecessary -- if all it's doing is serving as a list of Commands, it's just a wrapper around list, and so the code would be clearer if the classes that were using this just kept their own list and used map.  Is there an intent to put centralized logging/error handling logic here later?

ingest/expression_files/dense.py
```diff
@@ -0,0 +1,175 @@
...
+        # Represents row as an ordered dictionary
+        for row in self.csv_file:
+            numeric_scores = list(
+                map(lambda x: round(float(x.strip().strip("\'")), 3), row[1:])
+            )
+            valid_expression_scores = []
+            cells = []
+
+            for idx, expression_score in enumerate(numeric_scores, 1):
+                if expression_score > 0:
```
bistline, line 122: This should be `if expression_score != 0`.  In looking at the existing gene expression parsing code in the Ruby codebase, we only check if the value is _equal_ to zero, not greater.  While unusual, there are instances of negative values in dense matrices that have been normalized/processed in certain ways.

ingest/expression_files/expression_files.py
bistline: The structure of the objects yielded by both functions is the same - there are simply some values that get set by default that are different between the two.  Specifically the `name` of the array, and the `array_type` of either `expression` or `cells`.  I would support merging them together as well.

ingest/expression_files/expression_files.py
```diff
@@ -102,4 +113,46 @@ def set_data_array_gene_expression_values(
...
+        """
+        Load gene and data_array models into mongoDB
+        """
+        gene_doc_bulk_write_results = None
+        data_array_bulk_write_results = None
+        start_time = datetime.datetime.now()
+
+        # Creating Mongo bulk operations
+        data_array_bulk_operations = list(
+            map(lambda model: InsertOne(model), data_array_documents))
```
bistline, line 129: Is there a reason why we're doing a `list` of `InsertOne` operations and calling `bulk_write()`?  We're not doing a mix of write/update/delete operations, so it would stand to reason that calling `insert_many()` would suffice once the list of documents to insert reaches a certain pre-set limit.

ingest/expression_files/dense.py
```diff
@@ -0,0 +1,175 @@
...
+                    valid_expression_scores.append(expression_score)
+                    cells.append(self.header[idx])
+            gene = row[0]
+            if gene in self.gene_names:
+                raise ValueError(f'Duplicate gene: {gene}')
+            self.gene_names[gene] = True
+            print(f'Transforming gene :{gene}')
```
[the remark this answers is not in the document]
knapii-developments: It's was the ONLY thing that I could get persist on PAPI :( . Not saying that there isn't another way but I couldn't figure it out with python's logger

ingest/connection.py
```diff
@@ -0,0 +1,49 @@
...
+    def client(self, x):
+        if os.environ.get('DATABASE_HOST') is not None:
+            self._client = MongoClient(
+                host,
+                username=user,
+                password=password,
+                authSource=db_name,
+                authMechanism='SCRAM-SHA-1',
+            )
+        else:
```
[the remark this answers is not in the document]
knapii-developments: The [comment from ingest pipeline](https://github.com/broadinstitute/scp-ingest-pipeline/blob/ac8786dcf246ac8d34cd8bbe867c981e16c92c73/ingest/ingest_pipeline.py#L118) says that check is ' Needed to run tests in CircleCI.  TODO: add mock, remove this'.

ingest/expression_files/dense.py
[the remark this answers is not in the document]
eweitz: OK, I'm glad this information is getting persisted!

ingest/connection.py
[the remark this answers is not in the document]
eweitz: I wrote the original connection code and comment, so I can offer some context.

> Needed to run tests in CircleCI.

Tests in [`test_ingest.py`](https://github.com/broadinstitute/scp-ingest-pipeline/blob/master/tests/test_ingest.py) fail in CI without something like this, as is.  The intent of those tests is to verify observable output using the same input interface as PAPI (i.e. using `ingest_pipeline.py` CLI arguments).  Such tests are a fast and accurate way to verify code works after lower-level implementation overhauls -- e.g. after a big refactoring like this.

> TODO: add mock, remove this

The `else` in question here exists because we do not use a database mock library for MongoDB.  Lacking such a mock DB means we need to mock an implementation detail (i.e. this DB reference, and a `load` function), against the intent of `test_ingest.py`.

It would be better to integrate a MongoDB mock, like [mongomock](https://github.com/mongomock/mongomock).  We [had a similar mock for Firestore](https://github.com/broadinstitute/scp-ingest-pipeline/pull/26), but never got around to the same infrastructure for MongoDB after pivoting.

SCP-2000 tracks that test infrastructure goal.  Refining the comment to:

> // TODO (SCP-2000): Integrate MongoDB emulator to test_ingest.py, then remove this

would better convey that goal.

ingest/connection.py
```diff
@@ -0,0 +1,49 @@
+import abc
+import os
+
+from pymongo import MongoClient
+
+
```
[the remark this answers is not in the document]
knapii-developments: 43f97582fbe586a93e1619f3e96087a98797cc53

ingest/connection.py
```diff
@@ -0,0 +1,49 @@
...
+        self.user = os.environ['MONGODB_USERNAME']
+        self.password = os.environ['MONGODB_PASSWORD']
+        self.db_name = os.environ['DATABASE_NAME']
+
+    @property
+    def client(self):
+        return self._client
+
+    @client.setter
+    def client(self, x):
```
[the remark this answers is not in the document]
knapii-developments: This was removed as suggested by [Devon's comment](https://github.com/broadinstitute/scp-ingest-pipeline/pull/108/files#r447154638)

ingest/connection.py
```diff
@@ -0,0 +1,49 @@
...
+
+    @client.setter
+    @abc.abstractmethod
+    def client(self, x):
+        pass
+
+
+class MongoConnection(Connection):
+    """
+    A concrete class that defines a MongoDB client
```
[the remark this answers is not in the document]
knapii-developments: fbab66d5cc79235460d4eb3a7dcad4c73c4fcb64

ingest/connection.py
```diff
@@ -0,0 +1,49 @@
...
+        self.password = os.environ['MONGODB_PASSWORD']
+        self.db_name = os.environ['DATABASE_NAME']
+
+    @property
+    def client(self):
+        return self._client
+
+    @client.setter
+    def client(self, x):
+        if os.environ.get('DATABASE_HOST') is not None:
```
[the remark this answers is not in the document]
knapii-developments: fbab66d5cc79235460d4eb3a7dcad4c73c4fcb64

ingest/ingest_pipeline.py
knapii-developments: 40d0d5510a7af0b0b1dbc942b38444a986ff058f

ingest/ingest_pipeline.py
knapii-developments: 40d0d5510a7af0b0b1dbc942b38444a986ff058f

ingest/invoker.py
knapii-developments: e405fb2

ingest/expression_files/dense.py
```diff
@@ -0,0 +1,175 @@
...
+    from ingest_files import IngestFiles
+    from .monitor import trace
+    from connection import Connection
+
+except ImportError:
+    # Used when importing as external package, e.g. imports in single_cell_portal code
+    sys.path.append("../ingest")
+    from .expression_files import GeneExpression
+    from ingest_files import IngestFiles
+    from monitor import trace
```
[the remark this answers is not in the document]
knapii-developments: be99f49

ingest/expression_files/expression_files.py
knapii-developments: I don't think so. Updated in 5441f62

ingest/expression_files/dense_command.py
knapii-developments: This file was removed. Suggestion still applied in 
be99f49

tests/test_ingest.py
```diff
@@ -62,12 +63,26 @@ def mock_load(self, *args, **kwargs):
...
+
+    TODO:
+    Integrate MongoDB emulator for faster, higher-coverage tests (SCP-2000)
+
+    This will enable us to also verify (and thus cover) loading-code *outputs*,
+    unlike here where we merely give a way to verify loading-code *inputs*.
+    Doing so via integration tests will isolate us from implementation changes.
+    """
+    self.gene_docs = args[0]
+    self.data_arrary_docs = args[1]
```
eweitz, line 82:
```suggestion
    self.data_array_docs = args[1]
```

tests/test_ingest.py
```diff
@@ -102,121 +118,113 @@ def setup_ingest(self, args, mock_storage_client, mock_storage_blob):
...
-    #         'dense',
-    #     ]
-    #     ingest = self.setup_ingest(args)[0]
-    #
-    #     models = ingest.load_args[0]
-    #     for model in models:
-    #         # Ensure that 'ObjectID' in model is removed
-    #         del model['_id']
-    #         self.assertEqual(model, gene_models[model['name']])
+    def test_ingest_dense_matrix(self):
```
eweitz, line 123: I'm glad to see these tests working again!  Kudos.

ingest/ingestor.py
```diff
@@ -0,0 +1,19 @@
...
+    def execute_ingest(self, ):
+        pass
+
+    """
+    Returns true/false for whether this ingestor can handle the given file type
+    type should correspond to the
+    """
+    @abc.abstractmethod
+    def matches_file_type(self, type):
+        pass
```
eweitz, line 19: This doc should be on the line after the function signature, so IDEs can auto-detect it.  Also, the doc itself looks truncated at the end ("...corresponds to the" what?).

```suggestion
    @abc.abstractmethod
    def matches_file_type(self, type):
        """Returns true/false for whether this ingestor can handle the given file type
        type should correspond to the
        """
        pass
```

ingest/ingest_pipeline.py
```diff
@@ -148,12 +156,14 @@ def __init__(
...
     @my_debug_logger()
+    # Will be replaced by MongoConnection
     def get_mongo_db(self):
         host = os.environ['DATABASE_HOST']
         user = os.environ['MONGODB_USERNAME']
         password = os.environ['MONGODB_PASSWORD']
         db_name = os.environ['DATABASE_NAME']
-
+        print(f'db_name:{db_name},  user:{user}, host:{host}')
+        self.error_logger.error(f'db_name:{db_name},  user:{user}, host:{host}', extra=self.extra_log_params)
```
eweitz, line 166: Conventional spacing would enhance `grep`-ability.

```suggestion
        print(f'db_name: {db_name},  user: {user}, host: {host}')
        self.error_logger.error(f'db_name: {db_name},  user: {user}, host: {host}', extra=self.extra_log_params)
```

ingest/ingest_pipeline.py
```diff
@@ -116,10 +120,14 @@ def __init__(
...
         self.matrix_file = matrix_file
         self.matrix_file_type = matrix_file_type
+        self.extra_log_params = {'study_id': self.study_id, 'duration': None}
         if os.environ.get('DATABASE_HOST') is not None:
             # Needed to run tests in CircleCI.  TODO: add mock, remove this
+            self.error_logger.error(f"Database host is: {os.environ.get('DATABASE_HOST')}", extra=self.extra_log_params)
+            print(f"Database host is: {os.environ.get('DATABASE_HOST')}")
             self.db = self.get_mongo_db()
         else:
+            self.error_logger.error(f"Database host is: NONE", extra=self.extra_log_params)
```
eweitz, line 130: Having not-errors logged to an error log is confusing and seems likely to complicate debugging.

How about simply logging "info" and "errors" to the same file (perhaps "`run.log`" or some such)?  The SCP Rails app does this to good effect, and I think Cromwell does too.

ingest/ingest_pipeline.py
```diff
@@ -116,10 +120,14 @@ def __init__(
         self.study_file_id = study_file_id
         self.matrix_file = matrix_file
         self.matrix_file_type = matrix_file_type
+        self.extra_log_params = {'study_id': self.study_id, 'duration': None}
         if os.environ.get('DATABASE_HOST') is not None:
             # Needed to run tests in CircleCI.  TODO: add mock, remove this
```
eweitz, line 125: Closing the loop on [earlier feedback](https://github.com/broadinstitute/scp-ingest-pipeline/pull/108/files#r447150956):

```suggestion
            # Needed to run tests in CircleCI.
            # TODO (SCP-2000): Integrate MongoDB emulator to test_ingest.py, then remove this
```

ingest/expression_files/mtx.py
```diff
@@ -64,19 +64,16 @@ def __init__(self, mtx_path: str, study_file_id: str, study_id: str, **kwargs):
...
-        # pdb.set_trace()
-        yield from self.transform()
-        self.close()
+        for gene_docs, data_array_documents in self.transform():
+            self.load(gene_docs, data_array_documents)
+        return 0
 
     def extract_mtx(self, value):
         """
         Zgreps by gene index from mtx file to enhance performance/scale
```
eweitz, line 74: Reflecting your update in line 79:

```suggestion
        greps by gene index from mtx file to enhance performance/scale
```

Out of curiosity, what makes `grep` suitable but not `zgrep`?  (Do we know why `grep` performs better?)

ingest/expression_files/dense_ingestor.py
```diff
@@ -0,0 +1,179 @@
...
+            gene_models.append(self.Model(
+                {
+                    'name': formatted_gene_name,
+                    'searchable_name': formatted_gene_name.lower(),
+                    'study_file_id': self.study_file_id,
+                    'study_id': self.study_id,
+                    '_id': id,
+                    'gene_id': self.matrix_params['gene_id']
+                    if 'gene_id' in self.matrix_params
+                    else None,
```
eweitz, line 144: It's hard to read multi-line ternaries that are embedded in dictionaries.  Could you extract this into a simple `gene_id` variable?

ingest/expression_files/dense_ingestor.py
```diff
@@ -0,0 +1,179 @@
+"""Module for ingesting dense matrix files
+
+DESCRIPTION
+Module provides extract and transforms function for gene expression data for
+an dense matrix.
+
+PREREQUISITES
+Must have python 3.6 or higher.
```
eweitz, line 8: Noting prerequisites in `ingest_pipeline.py` makes sense; less so in modules that can't (or wouldn't) be run via CLI.

```suggestion
```

(This also reduces maintenance burden, e.g. we note the updated requirement for Python 3.7 in fewer places.)

ingest/expression_files/mtx.py
knapii-developments: zgrep is used for zipped files and was needed to ingest the largest matrix, a zipped file, in PAPI. When trying to fulfill [SCP-2577](https://broadworkbench.atlassian.net/browse/SCP-2577) is wasn't as performant as expected. The files we use for testing aren't zipped which is why I needed to revert zgrep to grep. I believe this method will become less relevant due to [SCP-2661](https://broadworkbench.atlassian.net/browse/SCP-2611#) and the fact that it wasn't performant.

ingest/expression_files/dense.py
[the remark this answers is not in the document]
knapii-developments: Just as a side note I think the [ticket](https://broadworkbench.atlassian.net/secure/RapidBoard.jspa?rapidView=37&projectKey=SCP&view=planning&selectedIssue=SCP-2525&issueLimit=100) you created will replace these logging statements? Is that not so?

ingest/expression_files/dense_ingestor.py
```diff
@@ -0,0 +1,179 @@
...
+            )
+            if len(valid_expression_scores) > 0:
+                print('ere')
+                for gene_cell_model in self.set_data_array_gene_cell_names(gene, id, cells):
+                    data_arrays.append(gene_cell_model)
+                for gene_expression_values in self.set_data_array_gene_expression_values(gene, id, valid_expression_scores):
+                    data_arrays.append(gene_expression_values)
+                if len(gene_models) == 5:
+                    num_processed += len(gene_models)
+                    print(f'Processed {num_processed} models, {str(datetime.datetime.now() - start_time)} elapsed')
```
devonbush, line 156: remove debugging print

ingest/expression_files/dense_ingestor.py
```diff
@@ -0,0 +1,179 @@
...
+                    self.error_logger.info(f'Processed {num_processed} models, {str(datetime.datetime.now() - start_time)} elapsed', extra=self.extra_log_params)
+                    yield (gene_models, data_arrays)
+                    gene_models = []
+                    data_arrays = []
+        yield (gene_models, data_arrays)
+        num_processed += len(gene_models)
+        print(f'Processed {num_processed} models, {str(datetime.datetime.now() - start_time)}')
```
devonbush, line 163: remove debugging print

ingest/expression_files/mtx.py
devonbush: cool, don't forget to update the comment

ingest/expression_files/mtx.py
```diff
@@ -23,7 +23,7 @@
 from bson.objectid import ObjectId
```
devonbush, line 24: let's exclude changes to this file from this commit, so we can focus our full attention on dense

ingest/ingest_pipeline.py
devonbush: ?

ingest/ingest_pipeline.py
devonbush: ?

ingest/expression_files/mtx.py
devonbush: cool, don't forget to update the comment

ingest/ingest_pipeline.py
devonbush: concur

tests/test_dense.py
```diff
@@ -1,101 +1,115 @@
...
 import sys
 from mock_data.dense_matrix_19_genes_100k_cells_txt.gene_models_0 import gene_models
 
 sys.path.append("../ingest")
-from dense import Dense
+from expression_files.mtx import MTXIngestor
+from expression_files.dense_ingestor import DenseIngestor
 
 
 class TestDense(unittest.TestCase):
```
devonbush, line 11: add comment for why these tests are commented out

ingest/connection.py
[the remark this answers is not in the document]
knapii-developments: f9d3268

ingest/ingestor.py
knapii-developments: Via [pep 257 -- Docstring Conventions](https://www.python.org/dev/peps/pep-0257/)  under [Multi-line Doc string](https://www.python.org/dev/peps/pep-0257/#id17) "The summary line may be on the same line as the opening quotes or on the next line."

eweitz: The issue here is having the docstring above the function signature -- the issue isn't the position of the summary line within the multiline docstring.  Your quoted clause from PEP 257 means this would be OK (albeit less common):

```
    @abc.abstractmethod
    def matches_file_type(self, type):
        """
        Returns true/false for whether this ingestor can handle the given file type
        type should correspond to the
        """
```

But no examples in PEP 257 put the docstring _above_ the signature -- so the suggested change is more conventional.  

That said, my comment was more motivated more by functional effect on developer experience than style compliance. 
 As is, the docstring cannot be auto-detected by IDEs.  The suggested change fixes that DX paper-cut.

ingest/expression_files/dense_ingestor.py
```diff
@@ -0,0 +1,187 @@
...
+    from .expression_files import GeneExpression
+    sys.path.append("../ingest")
+    from ..ingest_files import IngestFiles
+
+
+class DenseIngestor(GeneExpression, IngestFiles):
+    ALLOWED_FILE_TYPES = ["text/csv",
+                          "text/plain", "text/tab-separated-values"]
+
+    def __init__(self, file_path, study_file_id, study_id, **kwargs):
```
devonbush, line 32: Add a method comment.  And either rename the **kwargs param **matrix_params or add a comment explaining that kwargs will be used as matrix params

ingest/expression_files/dense_ingestor.py
```diff
@@ -0,0 +1,187 @@
...
+        GeneExpression.__init__(self, file_path, study_file_id, study_id)
+        IngestFiles.__init__(
+            self, file_path, allowed_file_types=self.ALLOWED_FILE_TYPES
+        )
+        self.matrix_params = kwargs
+        self.csv_file_handler, self.file_handler = self.open_file(
+            self.file_path)
+        self.gene_names = {}
+        self.header = next(self.csv_file_handler)
```
devonbush, line 41: Unless you have reason otherwise, I would move this statement to inside the execute_ingest method, since reading the header seems like more of an ingest-y thing to do

ingest/expression_files/dense_ingestor.py
```diff
@@ -0,0 +1,187 @@
...
+
+    def execute_ingest(self):
```
devonbush, line 43: Add a method comment.  This method can only ever be called once for a given instance of DenseIngestor, right?  That might be good to add in the comment

ingest/expression_files/dense_ingestor.py
```diff
@@ -0,0 +1,187 @@
...
+        # Row after header is needed for R format validation
+        row = next(self.csv_file_handler)
+        if DenseIngestor.is_valid_format(self.header, row):
+            # Reset csv reader to first gene row
+            self.csv_file_handler = self.open_file(self.file_path)[0]
+            next(self.csv_file_handler)
+            for gene_docs, data_array_documents in self.transform():
+                self.load(gene_docs, data_array_documents)
+        else:
+            raise ValueError("Dense matrix has invalid format")
```
devonbush, line 53: instead of using if/else, I would structure this block as
```
if not DenseIngestor.is_valid_format(self.header, row):
    raise ValueError(...)
self.csv_filehandler = ...
```

ingest/expression_files/dense_ingestor.py
```diff
@@ -0,0 +1,187 @@
...
+        for row in self.csv_file_handler:
+            numeric_scores = DenseIngestor.process_row(row)
+            valid_expression_scores, cells = DenseIngestor.filter_expression_scores(
+                numeric_scores, self.header)
+            cells = []
+            gene = row[0]
+            if gene in self.gene_names:
+                raise ValueError(f'Duplicate gene: {gene}')
+            self.gene_names[gene] = True
+            print(f'Transforming gene :{gene}')
```
devonbush, line 156: remove debugging print

ingest/expression_files/expression_files.py
```diff
@@ -109,3 +119,20 @@ def set_data_array_gene_expression_values(
             self.study_file_id,
         ).get_data_array():
             yield model
+
+    def load(self, gene_docs: List, data_array_docs: List):
+        start_time = datetime.datetime.now()
+        self.insert(gene_docs, self.COLLECTION_NAME)
+        self.insert(data_array_docs, 'data_array')
+        print(f'Time to load {len(gene_docs) + len(data_array_docs)} models: {str(datetime.datetime.now() - start_time)}')
```
devonbush, line 127: remove debugging print - use logger instead

ingest/expression_files/expression_files.py
```diff
@@ -109,3 +119,20 @@ def set_data_array_gene_expression_values(
...
+
+    def insert(self, docs: List, collection_name: str):
+        try:
+            self.mongo_connection._client[collection_name].insert_many(
+                docs,  ordered=False)
+        except BulkWriteError as bwe:
+            print(f"Error caused by inserting into collection '{collection_name}': {bwe.details}")
```
devonbush, line 134: remove debugging print - use logger instead

ingest/expression_files/expression_files.py
```diff
@@ -109,3 +119,20 @@ def set_data_array_gene_expression_values(
...
+            raise BulkWriteError(f'Error caused by data docs : {bwe.details}')
+        except Exception as e:
+            print(f"Error caused by inserting into collection '{collection_name}': {e}")
```
devonbush, line 137: remove debugging print - use logger instead if needed

ingest/expression_files/expression_files.py
```diff
@@ -109,3 +119,20 @@ def set_data_array_gene_expression_values(
...
+            raise Exception(f'{e}')
```
devonbush, line 138: You should raise the exception with the additional message you have above.  Alternatively, just do `raise e` so that you preserve the type of the root error

ingest/expression_files/mtx.py
devonbush: don't forget to update the comment

ingest/ingest_pipeline.py
```diff
@@ -140,20 +146,18 @@ def __init__(
...
                 "cluster", cluster_file)
-        if matrix_file is None:
-            self.matrix = matrix_file
         self.extra_log_params = {'study_id': self.study_id, 'duration': None}
         if subsample:
             self.cluster_file = cluster_file
             self.cell_metadata_file = cell_metadata_file
 
     @my_debug_logger()
+    # Will be replaced by MongoConnection
```
devonbush, line 155: The comment could use more detail.  When will it be replaced?  Why not now?  Is there a Jira ticket for that replacement

ingest/ingestor.py
```diff
@@ -0,0 +1,19 @@
...
+    action.
+    """
+    @abc.abstractmethod
+    def execute_ingest(self, ):
+        pass
+
+    """
+    Returns true/false for whether this ingestor can handle the given file type
+    type should correspond to the
+    """
```
devonbush, line 16: move comment to inside method and finish the sentence, per Eric's comment. (I think this might have been code from my merge -- sorry)

tests/test_dense.py
```diff
@@ -1,101 +1,129 @@
...
 sys.path.append("../ingest")
-from dense import Dense
+from expression_files.dense_ingestor import DenseIngestor
+from mock_data.dense_matrix_19_genes_100k_cells_txt.gene_models_0 import \
+    gene_models
 
 
 class TestDense(unittest.TestCase):
-    def test_validate_gene_keyword_false(self):
+
```
devonbush, line 12: Looking good.  I think we need to add tests that we handle single/double quotes in the header row (are there any other special characters we should strip from the headers?), and detect duplicates correctly even with quotes.

@jlchang  can you take a look at this file and let us know what are other expression file edge parsing cases we need to handle?

devonbush:
Jean pointed out that we should make sure that NaN values get filtered out during parsing as well (i.e. they should be treated as zeroes)

ingest/expression_files/dense_ingestor.py
```diff
@@ -0,0 +1,187 @@
...
+
+        Returns:
+            tuple (generator):
+                valid_expression_scores (list): non-zero gene scores
+                associated_cells (list): cells
+        """
+        associated_cells = []
+        valid_expression_scores = []
+        for idx, expression_score in enumerate(scores, 1):
+            if expression_score != 0:
```
jlchang, line 89: In anticipation of aggregated cell matrices (where "genes not assayed in one of the combined studies" get NaN values to indicate "not assayed") we should also filter out NaN as they can be treated as zero for visualization.

tests/test_dense.py
```diff
@@ -1,101 +1,129 @@
...
-        "Validates validate_gene_keyword() returns true correctly"
-        expression_matrix = Dense(
-            '../tests/data/dense_expression_matrix.txt',
-            '5d276a50421aa9117c982845',
-            '5dd5ae25421aa910a723a337',
-            tracer=None,
-        )
-        self.assertTrue(expression_matrix.validate_gene_keyword())
-        expression_matrix.close()
+        header = ["GENE", 'foo', 'foo2', 'foo3']
```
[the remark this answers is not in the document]
knapii-developments: I believe this is taken care of [here](https://github.com/broadinstitute/scp-ingest-pipeline/blob/3e6e92b888a063579686fe376e29c1445b6ffefd/tests/test_dense.py#L38)

tests/test_dense.py
knapii-developments: This is handled is this [test case](https://github.com/broadinstitute/scp-ingest-pipeline/blob/3e6e92b888a063579686fe376e29c1445b6ffefd/tests/test_dense.py#L35).

ingest/expression_files/dense_ingestor.py
knapii-developments: 4402ffd48b30f870f1f2edb04f3db548748a3b13

tests/test_dense.py
knapii-developments: eb08461

ingest/expression_files/dense_ingestor.py
knapii-developments: cf8b889

ingest/expression_files/expression_files.py
knapii-developments: a45ee4f

ingest/ingest_pipeline.py
knapii-developments: a68804f

ingest/ingestor.py
knapii-developments: This file wasn't suppose to be included. 12c83ea

ingest/expression_files/dense_ingestor.py
knapii-developments: Isn't that implied by having self as an argument?

ingest/expression_files/dense_ingestor.py
knapii-developments: aab53bd

ingest/expression_files/dense_ingestor.py
knapii-developments: aab53bd

tests/test_dense.py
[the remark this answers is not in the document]
jlchang: Ah, sorry I missed that!

ingest/expression_files/dense_ingestor.py
jlchang: Per [Devon's comment](https://github.com/broadinstitute/scp-ingest-pipeline/pull/108#issuecomment-666375511) filtering out NaN is something we do now in production. My comment was just indicating that there will be more NaN in future so it is best that we maintain parity with production now.

ingest/expression_files/dense_ingestor.py
```diff
@@ -0,0 +1,214 @@
...
+        return all(
+            [
+                DenseIngestor.has_unique_header(header),
+                DenseIngestor.has_gene_keyword(header, row),
+                DenseIngestor.header_has_valid_values(header),
+            ]
+        )
+
+    @staticmethod
+    def has_unique_header(header: List):
```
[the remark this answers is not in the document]
knapii-developments: This [ticket](SCP-2635 Ensure cell names from expression matrices are unique across files) tracks the issues

ingest/expression_files/expression_files.py
```diff
@@ -109,3 +117,25 @@ def set_data_array_gene_expression_values(
             self.study_file_id,
         ).get_data_array():
             yield model
+
+    def load(self, gene_docs: List, data_array_docs: List):
```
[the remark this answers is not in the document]
knapii-developments: Right now [we're doing 100k](https://github.com/broadinstitute/scp-ingest-pipeline/blob/1e77ac7994c7140b4c0f6ddab78b1112e1488c7b/ingest/ingest_files.py#L30). I'm not sure how this has effecting production.

ingest/expression_files/dense_ingestor.py
```diff
@@ -0,0 +1,214 @@
...
+            )
+            numeric_scores = DenseIngestor.process_row(valid_expression_scores)
+            cells = []
+            gene = row[0]
+            if gene in self.gene_names:
+                raise ValueError(f"Duplicate gene: {gene}")
+            self.gene_names[gene] = True
+            formatted_gene_name = DenseIngestor.format_gene_name(gene)
+            id = ObjectId()
+            if "gene_id" in self.matrix_params:
```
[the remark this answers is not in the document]
knapii-developments: This was just a misinterpretation of the models documentation. If dense doesn't take gene_ids as a passed in argument I'll remove it.

ingest/expression_files/dense_ingestor.py
devonbush: No, self just lets you know that this is a class method.  It doesn't let you know that this method is very tightly coupled to the current position of the csv_file_handler, such that calling this method twice would give bad results.  And it's just generally good practice at the start of long methods to have a comment explaining what the method is intended to do

ingest/expression_files/mtx.py
```diff
@@ -11,83 +11,77 @@
...
     def extract_mtx(self, value):
         """
-        Zgreps by gene index from mtx file to enhance performance/scale
+        Greps by gene index from mtx file to enhance performance/scale
         """
-        return subprocess.run(['zgrep', f'^{value}\s', self.mtx_local_path],
-                              stdout=subprocess.PIPE, stderr=subprocess.PIPE
-                              ).stdout.decode('utf-8')
+        return subprocess.run(
+            ["grep", f"^{value}\s", self.mtx_local_path],
```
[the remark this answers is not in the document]
knapii-developments: This implementation do longer applies due to resulting memory issues which are being addressed in this [ticket](https://broadworkbench.atlassian.net/browse/SCP-2577)

ingest/expression_files/mtx.py
```diff
@@ -116,21 +109,22 @@ def transform(self):
...
-                {
-                    'name': gene,
-                    'searchable_name': gene.lower(),
-                    'study_file_id': self.study_file_id,
-                    'study_id': self.study_id,
-                    'gene_id': gene_id,
-                    '_id': id,
-                }
-            ))
+            gene_id, gene = self.genes[int(mtx_gene_idx)].split("\t")
```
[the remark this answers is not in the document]
knapii-developments: This implementation was just to get MTX running to parse the 1M matrix in PAPI. It no longer applies. I'm working on MTX right now and will make note that we accept both csv and tsv files.

ingest/expression_files/mtx.py
```diff
@@ -11,83 +11,77 @@
...
-        yield from self.transform()
-        self.close()
+        for gene_docs, data_array_documents in self.transform():
+            self.load(gene_docs, data_array_documents)
+        return 0
 
     def extract_mtx(self, value):
         """
-        Zgreps by gene index from mtx file to enhance performance/scale
+        Greps by gene index from mtx file to enhance performance/scale
```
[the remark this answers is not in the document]
knapii-developments: See [comment](https://github.com/broadinstitute/scp-ingest-pipeline/pull/108#discussion_r465140056)

ingest/expression_files/dense_ingestor.py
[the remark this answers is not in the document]
devonbush: Hmm... yeah, that's tricky.  I would lean towards code just querying for the cell names based on the study_id that it knows.  It might require going down a couple of layers in query, but our python code has to know about those mongo structures since it's doing all the ingest that put those cell names there in the first place.

ingest/expression_files/expression_files.py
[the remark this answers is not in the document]
knapii-developments: We limit the amount of genes to [5/batch](https://github.com/broadinstitute/scp-ingest-pipeline/blob/a20fd6db1d69539aff10a1ba35c6e13fadcc9036/ingest/expression_files/dense_ingestor.py#L199). The corresponding data arrays will be included in that.

tests/test_dense.py
```diff
@@ -1,101 +1,157 @@
...
+        self.assertFalse(DenseIngestor.has_gene_keyword(invalid_header, row))
+        self.assertFalse(DenseIngestor.has_gene_keyword(empty_header, row))
+
+    def test_has_unique_header(self):
+        """Validates validate_unique_header() returns false correctly"""
+
+        header = ["GENE", "foo", "foo2", "foo3"]
+        invalid_header = ["GENE", "foo", "foo", "foo3"]
+        self.assertFalse(DenseIngestor.has_unique_header(invalid_header))
+        self.assertTrue(DenseIngestor.has_unique_header(header))
```
devonbush, line 57: Add a unit test here for NaN handling

knapii-developments: c2394bf

ingest/expression_files/dense_ingestor.py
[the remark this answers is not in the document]
jlchang: Thanks for creating a ticket to make sure that unique cells will be checked. As long as we have planned follow-on work, that satisfies my change request.

ingest/expression_files/dense.py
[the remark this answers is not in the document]
eweitz: For our reference: we verified this works.

ingest/expression_files/dense_ingestor.py
jlchang: Happy to see the new NaN handling!

ingest/ingest_pipeline.py
devonbush: remove the extra #

ingest/ingest_pipeline.py
devonbush: ?

tests/test_dense.py
```diff
@@ -1,101 +1,174 @@
...
+
+        self.assertTrue(DenseIngestor.header_has_valid_values(header))
+        self.assertFalse(DenseIngestor.header_has_valid_values(empty_value))
+        self.assertFalse(DenseIngestor.header_has_valid_values(space_value))
+        self.assertFalse(DenseIngestor.header_has_valid_values(nan_value))
+
+    def test_filter_expression_scores(self):
+        scores = ["BRCA1", 4, 0, 3, "0", None, "", "   ", "nan", "NaN", "Nan", "NAN"]
+        cells = ["foo", "foo2", "foo3", "foo4", "foo5", "foo6", "foo7"]
+        invalid_scores = ["BRCA1", 4, 0, 3, "0", None, "T"]
```
devonbush, line 32: moving this declaration of invalid_scores down to right above where it's used will make the test easier to read

knapii-developments: c5a6ac3

PR #111, opened 2020-07-07
Handle NA values in annotation files

knapii-developments:
Prior to the PR, there was no handling for NA values. By adding ```keep_default_na``` pandas will include the default NaN values when parsing the data and handle values as strings. By default the following values are interpreted as NaN: ‘’, ‘#N/A’, ‘#N/A N/A’, ‘#NA’, ‘-1.#IND’, ‘-1.#QNAN’, ‘-NaN’, ‘-nan’, ‘1.#IND’, ‘1.#QNAN’, ‘N/A’, ‘NA’, ‘NULL’, ‘NaN’, ‘n/a’, ‘nan’, ‘null’.

This PR satisfies [SCP-2545](https://broadworkbench.atlassian.net/browse/SCP-2545)


ingest/annotations.py
```diff
@@ -94,7 +94,8 @@ def create_data_frame(self):
             # multiIndex = pd.MultiIndex.from_tuples(index)
         index = pd.MultiIndex.from_tuples(columns)
         self.file = self.open_file(
-            self.file_path, open_as='dataframe', dtype=dtypes, names=index, skiprows=2
+            self.file_path, open_as='dataframe', dtype=dtypes, names=index, skiprows=2,
+            keep_default_na=False,
```
jlchang, line 98: This change prevents the [issue seen by Jon](https://broadinstitute.slack.com/archives/CSM9QUHMF/p1593097722253300) in SCP-2545 where NA in a group-based annotation in a cluster file is converted to NaN by ingest (via Pandas) - for group-based annotation, the annotation should remain a string. 

Eno and I chatted and there are outstanding questions about whether points with NA values should be subsampled and also if we're depending on Pandas detection of empty cells, we should at least add na_values with the empty string and find out which of NaN, nan, etc might be outputted by Seurat or ScanPy that we should be treating as NA values.

tests/test_annotations.py
```diff
@@ -54,18 +54,27 @@ def test_round(self):
...
                 # corrected testings of dataframe column dtype, using != always returns True
                 self.assertFalse(
-                    np.issubdtype(self.df.file[column].dtypes, np.number),
+                    np.issubdtype(cluster.file[column].dtypes, np.number),
                     "Group annotations must be string values",
                 )
+                # Each value in column should be treated as a string
+                for value in cluster.file[column]:
+                    self.assertTrue(
+                        type(value)== str,
```
eweitz, line 75: Linter seems to have missed:

```suggestion
                        type(value) == str,
```

PR #112, opened 2020-07-09
Turn off Pandas low_memory option (SCP-2214)

jlchang:
Incomplete coercion of numbers into strings for group-based annotations from [issue reported in Zendesk](https://broadinstitute.zendesk.com/agent/tickets/6039) was caused by "internal chunking" behavior that Pandas employs to reduce memory usage during file parsing. When this behavior is turned off (low_memory = False), coercion of numbers into strings is correctly performed. Note that regardless of low_memory setting, the entire file is read into memory (unless/until chunksize or iterator parameters are invoked) so this change does not affect overall memory usage of ingest pipeline.

This change was tested using PAPI in a local dev instance (thanks Devon!) to ensure that the largest metadata file in production could still be ingested after this change. 

This PR includes tests to check for complete coercion of numbers into strings for group-based annotations and a check for current behavior that an empty cell is converted to NaN and stored as NaN in Mongo (behavior that may change after work for SCP-2545)

This PR satisfies SCP-2214


tests/test_annotations.py
```diff
@@ -40,6 +40,32 @@ def setUp(self):
...
+        ), "numeric value should be coerced to string"
+        # This test will need to be adjusted when SCP-2545 is implemented so NA values
+        # become strings for group annotations. This test checks for current behavior,
+        # an empty cell becomes a NaN value which is a float.
+        assert isinstance(
+            lmtest.file['mixed_data']['group'][2], float
+        ), "expect empty cell conversion to NaN which is a float"
+        # numeric value in second chunk should still properly be coerced to string type
+        assert isinstance(
+            lmtest.file['mixed_data']['group'][32800], str
```
eweitz, line 66: Empty newlines between these blocks would help readability:

```suggestion
            lmtest.file['mixed_data']['group'][0], str
        ), "numeric value should be coerced to string"
        
        # This test will need to be adjusted when SCP-2545 is implemented so NA values
        # become strings for group annotations. This test checks for current behavior,
        # an empty cell becomes a NaN value which is a float.
        assert isinstance(
            lmtest.file['mixed_data']['group'][2], float
        ), "expect empty cell conversion to NaN which is a float"
        
        # numeric value in second chunk should still properly be coerced to string type
        assert isinstance(
            lmtest.file['mixed_data']['group'][32800], str
```

jlchang: Yes, thanks for the suggestion. `9457765`

PR #114, opened 2020-07-15
Coerce NaN to string for group annotations (SCP-2545)

jlchang:
Previously Ingest pipeline coercion of NaN to string for group annotations was incomplete. A corner case where strings and numbers are in a column with NaN values caused Pandas type inference allow NaN in those cases to not be coerced.

This PR forces all NaN in group annotations to be coerced to string, has metadata convention validation accept and includes tests to validate NaN coercion to string.

This PR sports SCP-2545.


ingest/annotations.py
```diff
@@ -256,3 +270,22 @@ def validate_format(self):
...
+    def validate_numeric_annots(self):
+        """Check numeric annotations are not string dtype
+        """
+        valid = True
+        for annot_header in self.file.columns[0:]:
+            annot_name = annot_header[0]
+            annot_type = annot_header[1]
+            if (
+                annot_type == 'numeric'
+                and self.file[annot_name].dtypes[annot_type] == 'O'
```
bistline, line 283: Are we sure this is going to work reproducibly?  I wasn't sure what this line was checking, so I looked at the `pandas` docs, and then I tried some local testing.  Instead of getting back `'0'` for a string column, I get back `dtype('0')`, which is a `numpy.dtype`.

Example:

```
>>> d = {'col1': [1, 2], 'col2': [3, 4], 'col3': ['foo', 'bar']}
>>> df = pd.DataFrame(data=d)
col1  col2 col3
0     1     3  foo
1     2     4  bar
>>> df.dtypes
col1     int64
col2     int64
col3    object
dtype: object
>>> df.dtypes['col3']
dtype('O')
```

ingest/annotations.py
```diff
@@ -9,27 +9,38 @@
...
+        __name__ + '_errors', 'errors.txt', level=logging.ERROR
+    )
+
     def __init__(
         self, file_path, allowed_file_types, study_id=None, study_file_id=None
     ):
         IngestFiles.__init__(self, file_path, allowed_file_types)
         if study_id is not None:
             self.study_id = ObjectId(study_id)
+            self.extra_log_params = {'study_id': self.study_id, 'duration': None}
```
eweitz, line 41: Worth noting, but not urgent: these event properties are subtly incompatible with Mixpanel analytics.  Mixpanel uses:

* camelCase instead of snake_case
* `perfTime` and not `duration` (nor `latency`) to track time intervals like this
* Study accession instead of study ID

These logs are not yet sent to Mixpanel, but if/when they are, they would ideally be compatible with SCP logging from other sources (client-side JS, server-side Ruby).

ingest/annotations.py
```diff
@@ -98,12 +109,19 @@ def create_data_frame(self):
...
+        if 'group' in self.annot_types:
+            group_columns = self.file.xs(
+                "group", axis=1, level=1, drop_level=False
+            ).columns.tolist()
+            try:
+                # coerce group annotations to type string
+                self.file[group_columns] = self.file[group_columns].astype(str)
+            except Exception as e:
+                self.error_logger.error(
+                    "Unable to coerce group annotation to string type",
```
eweitz, line 121: It's odd that our auto-formatter permits inconsistent quotes -- double here, single in line 112.

ingest/annotations.py
jlchang: The type you're getting back is a capital O for "object". I've checked and the conditional works if I check for 'object' rather than 'O' so I've updated the function for clarity. The line is checking that the study owner has declared the metadata to be a numeric type but pandas dtype inference has decided it can only be an object type [pandas order of type checking in inference, int -> float (-> bool) -> object]

ingest/annotations.py
jlchang: created SCP-2591 to track this concern.

ingest/annotations.py
jlchang: Sadly Black would enforce using all double quotes. Black is generally not configurable. This particular choice is problematic enough that Black has a --skip-string-normalization as an option, which we use. So Black is allowing inconsistency unless we want enforcement of all double quotes.

eweitz: Could we use single-quotes here for consistency, in that case, or would they trigger a warning / be reformatted?

jlchang: Sure, I've gone through the file and tried to fix all the quoting inconsistencies `22cec53`. It's a pity there's no way to have Black enforce single quotes.

PR #115, opened 2020-07-17
add ontology_browser_url, free text cell_type__custom (SCP-2566)

jlchang:
Removing controlled list constraint on cell_type__custom to increase flexibility for U19 group's metadata upload process.

Adding ontology_browser_url to metadata convention to provide URL for browser link in faceted search UI (SCP-2562)

This PR satisfies SCP-2566


schema/README.md
```diff
@@ -21,21 +21,16 @@
...
-
-* Test by running:
-```
-pytest -k test_external_metadata_convention
 ```  
 
 Notes:
 * Tests in test_validate_metadata.py use current metadata convention (except for invalid metadata convention test)
 
+* Specifically test_bigquery_json_content is expected to fail with the metadata convention is updated. The reference file, bq_test.json, must be updated (replace existing file with the generated addedfeed000000000000000.json file)
```
eweitz, line 29:
```suggestion
* Specifically test_bigquery_json_content is expected to fail when the metadata convention is updated. The reference file, bq_test.json, must be updated (replace existing file with the generated addedfeed000000000000000.json file)
```

scripts/serialize_convention.py
```diff
@@ -187,18 +189,31 @@ def set_file_names(project, version):
...
+        return True
     else:
         return None
 
 
+def check_urls(urls, attribute):
+    """Check if URL is valid
+    """
+    if len(urls) > 1:
+        print(f"Checking {len(urls)} urls for {attribute}")
```
eweitz, line 208:
```suggestion
        print(f"Checking {len(urls)} URLs for {attribute}")
```

PR #119, opened 2020-07-27
#SCP-2552 Implement expression file validation for dense matrices

knapii-developments:
This PR restores format validation and tests for dense matrices. 

This ticket satisfies [SCP-2552](https://broadworkbench.atlassian.net/browse/SCP-2552)


ingest/expression_files/dense_ingestor.py
```diff
@@ -44,62 +44,47 @@ def __init__(self, file_path, study_file_id, study_id, **kwargs):
...
+            self.error_logger.error(
+                "Duplicate header values are not allowed", extra=self.extra_log_params
+            )
+            return False
+        return True
+
+    def has_gene_keyword(self):
+        """Validates that 'Gene' is the first value in header"""
+        # File is an R formatted file
+        if (self.header[-1] == '') and (self.header[0].upper() != 'GENE'):
```
bistline, line 76: In `R`, if a user outputs a CSV using `write.table()`, what you get is a header line that is one entry shorter than each successive line.  Specifically - there is no padding for the `GENE` keyword.  That's what we want to check for here.

knapii-developments: Fixed in 15dc741

knapii-developments:
> Looks good!
> 
> Like validation functions for metadata files are in `ingest/validation/validate_metadata.py`, I wonder if similarly separating validation functions for expression files into another module in `ingest/validation/` would make sense. This would promote internal consistency, and ease reuse of these functions in `manage_study.py`, our public CLI.
> 
> That suggestion is out of scope here, but perhaps worth considering for later.

Format validations exist within each class. I wonder if validate_metadata.py was isolated because particular to metadata convention. The name isn't the clearest. 
I agree with your second point about mange_study.py. This [ticket](https://broadworkbench.atlassian.net/browse/SCP-2621) tracks it

PR #124, opened 2020-08-10
# SCP-2635 Ensure cell names from expression matrices are unique across files

knapii-developments:
This PR adds a validation that checks if an expression file has unique cells by querying MongoDB.


ingest/expression_files/expression_files.py
```diff
@@ -64,6 +64,41 @@ def set_dataArray(self):
...
+        )
+        # Query did not return results
+        if not query_results:
+            return True
+        # Flatten query results
+        existing_cells = [
+            values
+            for cell_values in query_results
+            for values in cell_values.get("values")
+        ]
```
devonbush, line 99: we're going to want the error that we provide to the user to contain at least one of the duplicate cell names, right?  So I suggest this method also return a tuple of boolean/string.  (And only return the first few duplicate names if there are too many to print.  e.g.  "Found X duplicate cells corresponding to already existing expression matrices, including foo1, foo2..."

ingest/expression_files/dense_ingestor.py
```diff
@@ -40,21 +40,36 @@ def __init__(self, file_path, study_file_id, study_id, **matrix_kwargs):
...
         # dependency on position in text file.
         # Row after header is needed for R format validation
         row = next(self.csv_file_handler)
-        if not DenseIngestor.is_valid_format(self.header, row):
+        if not DenseIngestor.is_valid(
+            self.header,
+            row,
+            query_params=(self.study_id, self.mongo_connection._client),
+        ):
             raise ValueError("Dense matrix has invalid format")
```
devonbush, line 56: Is there any way for the user to know what the specific issue with the file was?  Or will they get the same error message for both a malformed header and a duplicate gene name?

ingest/expression_files/dense_ingestor.py
```diff
@@ -40,21 +40,36 @@ def __init__(self, file_path, study_file_id, study_id, **matrix_kwargs):
...
-    def matches_file_type(file_type):
-        return file_type == "dense"
+    @staticmethod
+    def is_valid(header, row, query_params):
+        return all(
+            [
+                DenseIngestor.has_unique_header(header),
+                DenseIngestor.has_gene_keyword(header, row),
+                DenseIngestor.header_has_valid_values(header),
+                GeneExpression.has_unique_cells(header, *query_params),
```
devonbush, line 70: this method needs a way to return the specific problem -- I recommend changing the method so it returns a tuple of [boolean, string], where the boolean is true/false for valid /invalid, and the string is the error message if invalid

tests/test_dense.py
```diff
@@ -88,54 +87,83 @@ def test_duplicate_gene(self):
...
     @patch(
         "expression_files.dense_ingestor.DenseIngestor.transform",
         return_value=[("foo1", "foo2")],
     )
-    def test_execute_ingest(self, mock_load, mock_transform):
+    @patch(
+        "expression_files.expression_files.GeneExpression.has_unique_cells",
+        return_value=True,
+    )
+    def test_execute_ingest(self, mock_load, mock_transform, mock_has_unique_cells):
```
devonbush, line 166: let's add a unit test where you mock the mongo client .find method, so you can actually test the logic of the has_unique_cells  method, and the error message it generates

ingest/expression_files/expression_files.py
```diff
@@ -64,6 +64,41 @@ def set_dataArray(self):
...
+        return not any(name in existing_cells for name in cell_names)
```
devonbush, line 100: I would use set intersection here so you can identify the total number and first few duplicates. set(lst1) & set(lst2)

ingest/expression_files/expression_files.py
```diff
@@ -64,6 +64,41 @@ def set_dataArray(self):
...
+
+         Parameters:
+            cell_names (List[str]): List of cell names in matrix
+            study_id (ObjectId): The study id the cell names belong to
+            client : MongoDB client
+        """
+        COLLECTION_NAME = "data_arrays"
+        query = {
+            "$and": [
+                {"linear_data_type": "Study"},
```
bistline, line 80: Not strictly necessary, but you could add `{"linear_data_id": study_id}`.

ingest/expression_files/expression_files.py
bistline, line 100: Nice!

ingest/expression_files/expression_files.py
knapii-developments: Tracked [here](https://broadworkbench.atlassian.net/browse/SCP-2660)

ingest/expression_files/dense_ingestor.py
knapii-developments: Tracked [here](https://broadworkbench.atlassian.net/browse/SCP-2660)

ingest/expression_files/dense_ingestor.py
knapii-developments: Tracked [here](https://broadworkbench.atlassian.net/browse/SCP-2660)

tests/test_ingest.py
```diff
@@ -40,7 +40,7 @@
 from mock_data.matrix_mtx.gene_model_0 import expected_model
 from gcp_mocks import mock_storage_client, mock_storage_blob
 
-sys.path.append('../ingest')
+sys.path.append("../ingest")
```
eweitz, line 43: Quote noise like this makes it hard to see relevant differences.  

It could be avoided by using `--skip-string-normalization` in one's IDE, as in [our Git hook](https://github.com/broadinstitute/scp-ingest-pipeline/blob/master/.pre-commit-config.yaml#L12).

Or we could remove that Black argument from our hook and our IDE configurations, and double-quote all strings in Python.  This would increase our Python's difference from our JavaScript, but ensuring readable diffs with minimal IDE configuration seems more important.

Eno, Jean, and I chatted and we lean toward removing `--skip-string-normalization`.

ingest/expression_files/expression_files.py
```diff
@@ -64,6 +64,41 @@ def set_dataArray(self):
...
 
+    @staticmethod
+    def has_unique_cells(cell_names: List, study_id, client):
+        """Checks cell names against database to confirm matrix contains unique
+            cell names
+
+         Parameters:
+            cell_names (List[str]): List of cell names in matrix
+            study_id (ObjectId): The study id the cell names belong to
+            client : MongoDB client
```
eweitz, line 75:
```suggestion
            client: MongoDB client
```

ingest/expression_files/expression_files.py
knapii-developments: I believe you took care of this

ingest/expression_files/expression_files.py
knapii-developments: You took care of it :D

ingest/expression_files/expression_files.py
knapii-developments: Im adding this suggestion in another PR.

PR #127, opened 2020-08-11
# SCP 2657 Add checks for R formatted files for dense matrix file format validations

knapii-developments:
The transform function did not take into account r formatted files. An index/out of range error would've occurred due to the length header being one less than the other rows.


ingest/expression_files/dense_ingestor.py
```diff
@@ -191,15 +203,18 @@ def transform(self):
         num_processed = 0
         gene_models = []
         data_arrays = []
-        for all_cell_model in self.set_data_array_cells(self.header[1:], ObjectId()):
+        # Cell names are formatted differently in R files
+        cell_names = self.header if self.is_r_file else self.header[1:]
```
bistline, line 207: Like this enhancement - it makes `filter_expression_scores` much more regular.

tests/test_dense.py
```diff
@@ -34,12 +34,12 @@ def test_check_header_valid_values(self):
...
     def test_filter_expression_scores(self):
         scores = ["BRCA1", 4, 0, 3, "0", None, "", "   ", "nan", "NaN", "Nan", "NAN"]
-        cells = ["foo", "foo2", "foo3", "foo4", "foo5", "foo6", "foo7"]
+        cells = ["foo", "foo2", "foo3", "foo4", "foo5"]
         actual_filtered_values, actual_filtered_cells = DenseIngestor.filter_expression_scores(
             scores[1:], cells
         )
         self.assertEqual([4, 3], actual_filtered_values)
-        self.assertEqual(["foo2", "foo4"], actual_filtered_cells)
+        self.assertEqual(["foo", "foo3"], actual_filtered_cells)
```
jlchang, line 42: The test here should have another valid value after all the values that are expected to be filtered to demonstrate that filtered values and cells names do not get "out of sync" during filtering because of the non-numeric NaN values.

PR #129, opened 2020-08-11
Harden Genomes Pipeline and make_toy_data.py (SCP-2491)

eweitz:
This improves robustness and maintainability for the SCP Genomes Pipeline and make_toy_data.py.  

Previously, these two pieces of infrastructure had no automated tests, and a tedious complex of parameters.  The code had moved from another repo in #102 to reduce technical debt after gaining significant features, but those two liabilities remained.

Now those risks are greatly reduced.  Genomes Pipeline (`genomes_pipeline.py`, previously named `process_genome_references.py`) now has an integration test in `test_genomes.py` and `make_toy_data.py` has one in `test_toy_data.py`.  To enable tests, the code has been restructured to enable use as a Python module.  Altogether, this substantially increases `scp-ingest-pipeline` test coverage -- from [52%](https://codecov.io/gh/broadinstitute/scp-ingest-pipeline/commit/d8454bb79a72ad310ca4c56d8bf6043405d4be26) before to [67%](https://codecov.io/gh/broadinstitute/scp-ingest-pipeline/tree/579edd813aad5bb07bc17b4ae9a76ce633a45ded) after.

Also, Genomes Pipeline is now more object-oriented, to simplify stateful transformations of genome reference data.  Static analysis via flake8 has also been restored for those files.  These two changes improve maintainability for four modules comprising ~700 lines of code, which is about a quarter of the code in `scp-ingest-pipeline`.

While manually verifying this work, I discovered that reference data can no longer be uploaded to GCS due to network timeouts.  This seems to be due to slower networks, and independent of these changes.  SCP-2662 tracks that.

This satisfies SCP-2491.


eweitz:
Oddly, even though I used e.g. `git mv ingest/genomes/process_genome_references.py ingest/genomes/genomes_pipeline.py`, the PR diff shows crude renamings.  This means Git history is lost for those files.  

The affected histories are short, as those files were recently migrating from another repo.  So unless anyone feels strongly I'll plan to not patch them.

ingest/genomes/genome_annotation_metadata.py
```diff
@@ -37,17 +33,17 @@ def upload_ensembl_gtf_products(ensembl_metadata, scp_species, config):
     remote_output_dir = config['remote_output_dir']
     remote_prod_dir = config['remote_prod_dir']
 
-    storage_client = get_gcs_storage_client(vault_path)
+    storage_client = utils.get_gcs_storage_client(vault_path)
```
devonbush, line 36: nice, I like how you made these easier to read where the functions come from

ingest/genomes/genome_annotation_metadata.py
```diff
@@ -37,17 +33,17 @@ def upload_ensembl_gtf_products(ensembl_metadata, scp_species, config):
...
 
     # FireCloud workspace for canonical SCP reference data:
     # https://portal.firecloud.org/#workspaces/single-cell-portal/scp-reference-data
     reference_data_bucket = gcs_bucket
-    target_folder = remote_output_dir
+    # target_folder = remote_output_dir
```
devonbush, line 41: either remove, or add comment explaining why this might need to be uncommented in the future

ingest/genomes/genome_annotation_metadata.py
```diff
@@ -128,12 +124,12 @@ def update_meta_row(row, org_metadata, annot_metadata):
...
 
 
-def record_annotation_metadata(ensembl_metadata, scp_species):
+def record_annotation_metadata(output_dir, ensembl_metadata, scp_species):
     """Write annotation URLs, etc. to species metadata reference TSV file
     """
     new_metadata_ref = []
 
-    ref_file = 'species_metadata_reference.tsv'
+    ref_file = f'{output_dir}species_metadata_reference.tsv'
```
devonbush, line 132: is this a place where os.path.join might be more robust? 
 https://docs.python.org/3/library/os.path.html

tests/test_genomes.py
```diff
@@ -0,0 +1,87 @@
...
+
+# Run all tests and see print() output
+pytest -s
+
+# Run only tests in test_genomes.py
+pytest test_genomes.py
+
+# Run all tests, show code coverage metrics
+pytest --cov=../ingest/
+
```
devonbush, line 27: I think it's overkill to be repeating pytest usage documentation here.    Maybe just add a link to the pytest docs from the main readme (line 76ish)?

tests/test_make_toy.py
```diff
@@ -0,0 +1,45 @@
...
+pytest
+
+# Run all tests and see print() output
+pytest -s
+
+# Run only tests in test_genomes.py
+pytest test_make_toy.py
+
+# Run all tests, show code coverage metrics
+pytest --cov=../ingest/
```
devonbush, line 20: same, I don't think it's useful to repeat pytest docs at the top of every test

ingest/genomes/genome_annotation_metadata.py
eweitz: Good point, I add a relevant TODO comment in 8a486fe.

ingest/genomes/genome_annotation_metadata.py
eweitz: Yup!  Done.  It's more verbose, but trading that for robustness makes sense.

tests/test_genomes.py
eweitz: Yeah, I can see how repeating these examples in each integration test module is a bit much.  At the same time, I think only linking to pytest docs makes conveying tailored examples too inconvenient.

So in 8a486fe I've consolidated this handful of conveniences into the "Test" section of the README, added a pointer here, and removed the verbose examples from the tests themselves.

tests/test_make_toy.py
eweitz: Agreed; details are in my reply over there.

PR #130, opened 2020-08-13
fixing ingest empty batch error (SCP-2669)

devonbush:


ingest/expression_files/dense_ingestor.py
```diff
@@ -28,6 +28,8 @@
 class DenseIngestor(GeneExpression, IngestFiles):
     # ToDo SCP-2635
     ALLOWED_FILE_TYPES = ["text/csv", "text/plain", "text/tab-separated-values"]
+    GENE_BATCH_SIZE = 5
+    DATA_ARRAY_BATCH_SIZE = 1000
```
jlchang, line 32: Devon, thanks for explaining the intent of `DATA_ARRAY_BATCH_SIZE`. Because this constant isn't intended to limit the number of data arrays that get loaded, it was helpful for me to understand how it was being used.

ingest/expression_files/dense_ingestor.py
```diff
@@ -28,6 +28,8 @@
 class DenseIngestor(GeneExpression, IngestFiles):
     # ToDo SCP-2635
     ALLOWED_FILE_TYPES = ["text/csv", "text/plain", "text/tab-separated-values"]
+    GENE_BATCH_SIZE = 5
```
bistline, line 31: Out of curiosity, why is `GENE_BATCH_SIZE` so much lower than `DATA_ARRAY_BATCH_SIZE` when the `Gene` documents are so much smaller overall compared to `DataArray` documents?

knapii-developments, line 31: I don't think we should be batching by gene. The gene documents are very small. Batching by gene will make more calls to the database than necessary. If this is kept the amount of genes should be significantly larger than the amount of data arrays.

bistline: +1

tests/test_dense.py
```diff
@@ -177,3 +177,15 @@ def test_transform_fn(self):
...
+        Assures transform function creates gene data model correctly for files
+        with a number of genes that is a multiple of GENE_BATCH_SIZE (SCP-2669)
+        """
+        expression_matrix = DenseIngestor(
+            "../tests/data/dense_matrix_10_genes_15_cells.txt",
+            "5d276a50421aa9117c982845",
+            "5dd5ae25421aa910a723a337",
+        )
+        for actual_gene_models, actual_data_arrays in expression_matrix.transform():
+            self.assertEqual(DenseIngestor.GENE_BATCH_SIZE, len(actual_gene_models))
```
knapii-developments, line 191: I think this test should check that a total of 'x' cells (an amount equal to the amount of lines in the file -1) were processed as well that the batch size is **less than or equal** to the batch size. As we run into edge cases, these test files are updated to reflect the case. If someone adds additional genes or the amount of genes is not a multiple of 5 this test will fail.

ingest/expression_files/dense_ingestor.py
```diff
@@ -28,6 +28,8 @@
 class DenseIngestor(GeneExpression, IngestFiles):
     # ToDo SCP-2635
     ALLOWED_FILE_TYPES = ["text/csv", "text/plain", "text/tab-separated-values"]
```
knapii-developments, line 30: Batch sizes values aren't particular to dense matrices but to all expression data. I think these values should move to the GeneExpression class.

ingest/expression_files/dense_ingestor.py
```diff
@@ -228,19 +230,26 @@ def transform(self):
...
+                    f"{str(datetime.datetime.now() - start_time)} elapsed",
+                    extra=self.extra_log_params,
+                )
+                yield gene_models, data_arrays
+                gene_models = []
+                data_arrays = []
+
+        # load any remaining models (this is necessary here since there isn't an easy way to detect the
+        # last line of the file in the iteration above
+        if len(gene_models) > 0:
```
knapii-developments, line 249: Non blocker but this can be ``` if gene_model```

tests/test_dense.py
devonbush: The strict equality is actually really important here.  If someone adds more genes to that file (so, e.g. it has 11 genes), it won't actually be testing what we want to test). As the comment says, this is for testing `files with a number of genes that is a multiple of GENE_BATCH_SIZE `  So a failure in this test would correctly tell someone they shouldn't have added genes to that file

ingest/expression_files/dense_ingestor.py
devonbush: Great question -- that limit on 5 genes was there from before.  I suspect that was a limit we used as a quick-and-dirty batching size when we made the first proof-of-concept of this ingestor, and then just stuck around...

tests/test_dense.py
knapii-developments: In that event we should still check for the first case scenario. And a comment should be added that the file should not be touched.

devonbush: I think what's important for test files is that the name includes the things that are important about the file -- e.g. 10_genes_15_cells  or error_headers.  If you're changing something that requires a name change to the file, that tells you you're changing something a test relies on.  So I'm not worried about this file being touched any more than I'd be worried about, e.g. someone updating the invalid_array_v2.1.2.txt file to not have any arrays.  Someone would have to both ignore the file name AND a failing test in order to make that change.

ingest/expression_files/dense_ingestor.py
devonbush: I can see this going either way -- `if gene_model` might be more pythonic, but in critical parsing code I'd rather be really explicit about what we are testing.  it's too easy to forget what counts as truthy in Python/Ruby/Javascript (e.g. in javascript `if ([])` evaluates as true) and have bugs creep in as a result

ingest/expression_files/dense_ingestor.py
devonbush: agree

PR #135, opened 2020-08-18
Script for ingest stress test (SCP-2649)

jlchang:
`ingest_smoke_test.py` will create a study and upload a dense matrix, cluster and metadata file to the study using manage-study (an installed executable from the single-cell-portal python package). By default this is run against the SCP staging server.

Setup:
- "smoke" virtualenv activated with single-cell-portal installed (see [setup instructions](https://docs.google.com/document/d/11BZQKstFgENqGIHaFuOOrHpYOedXXS86JUt1bNngTdk/edit#heading=h.f4m0pwfbcd7a))
- VPN (to access staging environment)
- have correct Google identity (non-SA account)
- Access token  ``ACCESS_TOKEN=`gcloud auth print-access-token` ``

Run a smoke test:
```
python ingest_smoke_test.py --token=$ACCESS_TOKEN
```

This supports SCP-2649.


scripts/ingest_smoke_test.py
```diff
@@ -0,0 +1,232 @@
...
+# Usage conditions: must have VPN access to staging server
+#                   files being uploaded cannot already exist in createTest
+python ingest_smoke_test.py --token=$ACCESS_TOKEN --dev-run
+"""
+
+import argparse
+import subprocess
+from datetime import datetime
+
+DATA = {
```
bistline, line 25: While this is great for our purposes, I'm hesitant to hard code these files in a public repo, since I'm not sure all of this data is public.  I'm fairly certain the data from the staging studies is not (it may not have ever been published, so I'm not sure where that leaves us).

scripts/ingest_smoke_test.py
```diff
@@ -0,0 +1,232 @@
...
+        '--number',
+        help='number of studies to create for stress test',
+        type=int,
+        default=1,
+    )
+
+    parser.add_argument(
+        '--environment',
+        help='server to test against [development,staging,production]',
+        default='staging',
```
eweitz, line 68: As in our [public CLI](https://github.com/broadinstitute/single_cell_portal/blob/master/scripts/cli_parser.py#L50):

```suggestion
        help='server to test against',
        choices=['development', 'staging', 'production'],
        default='staging',
```

Using `choices` gives built-in validation, and includes the options in `--help` output.

scripts/ingest_smoke_test.py
```diff
@@ -0,0 +1,232 @@
...
+    parser.add_argument(
+        '--dev-run',
+        action='store_true',
+        help='Run using stress_20200810-1357030 as study.',
+    )
+
+    parser.add_argument(
+        '--data-set',
+        default='small',
+        help='Select data set for stress test [small, medium, large].',
```
eweitz, line 86:
```suggestion
        help='Select data set for stress test.',
        choices=['small', 'medium', 'large']
```

scripts/ingest_smoke_test.py
```diff
@@ -0,0 +1,232 @@
...
+
+    parser.add_argument(
+        '--environment',
+        help='server to test against [development,staging,production]',
+        default='staging',
+    )
+
+    parser.add_argument(
+        '--token',
+        default=None,
```
eweitz, line 73: Unlike the public CLI, use of this script always requires a token, so:

```suggestion
        required=True,
```

This would also ensure an error is thrown if this required field is missing:

scripts/ingest_smoke_test.py
```diff
@@ -0,0 +1,232 @@
...
+
+DESCRIPTION
+Launch single study creation followed by file upload using manage-study
+OR concurrent ingest requests for X studies (stress test)
+
+PREREQUISITES
+set up access token on command line
+ACCESS_TOKEN=`gcloud auth print-access-token`
+
+EXAMPLE
```
eweitz, line 11:
```suggestion
EXAMPLES
```

scripts/ingest_smoke_test.py
devonbush: As long as these buckets are private (and it seemed to require me to login to access them), I don't think using them for testing is materially different than having the data live on our staging servers.  However, I might suggest reaming the files as something generic like "scp-large-test-[x].tsv", to address the concern that we're expressing favortism to certain studies/groups in our source, and then putting a README in the google bucket that describes where the files come from

scripts/ingest_smoke_test.py
```diff
@@ -0,0 +1,232 @@
+"""Launch ingest pipeline smoke test
+
+DESCRIPTION
+Launch single study creation followed by file upload using manage-study
+OR concurrent ingest requests for X studies (stress test)
+
+PREREQUISITES
+set up access token on command line
+ACCESS_TOKEN=`gcloud auth print-access-token`
```
devonbush, line 9: I might add a note, since this is a public repo, that this is an internal-only script for SCP developers since it relies on access to a non-public bucket

scripts/ingest_smoke_test.py
jlchang: Agreed. Files renamed `ef57bde`

scripts/ingest_smoke_test.py
jlchang: note added `ef57bde`

scripts/ingest_smoke_test.py
```diff
@@ -0,0 +1,232 @@
...
+
+    msaction = ManageStudyAction(args.data_set)
+
+    # generate study names
+    test_id = datetime.now().strftime('%Y%m%d-%H%M%S')
+    if args.number == 1:
+        test_type = "smoke"
+    else:
+        test_type = "stress"
+    study_names = [test_type + '-' + test_id + '-' + str(i) for i in range(args.number)]
```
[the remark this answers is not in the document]
jlchang: Makes sense, I support readability (and consistency is good too). Updated to f-string in `ef57bde`

scripts/ingest_smoke_test.py
bistline: That would assuage my concerns

scripts/ingest_smoke_test.py
```diff
@@ -0,0 +1,232 @@
...
+            "upload_dense_expression": [
+                'upload-expression',
+                '--study-name',
+                study,
+                '--file',
+                DATA[self.data_set]['dense'],
+                '--species',
+                "human",
+                '--genome',
+                "hg19",
```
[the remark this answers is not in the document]
jlchang: updated genome value to use the default genome name `38d94ae`
created SCP-2678 to follow up on what "hg19" didn't cause ingest to fail

scripts/ingest_smoke_test.py
jlchang: New to this feature, love it!  `38d94ae`

scripts/ingest_smoke_test.py
jlchang: `38d94ae`

scripts/ingest_smoke_test.py
jlchang: Good catch!  `38d94ae`

scripts/ingest_smoke_test.py
jlchang: `38d94ae`

scripts/ingest_smoke_test.py
```diff
@@ -0,0 +1,232 @@
...
+    if args.number == 1:
+        test_type = "smoke"
+    else:
+        test_type = "stress"
+    study_names = [test_type + '-' + test_id + '-' + str(i) for i in range(args.number)]
+
+    # create studies unless it is a dev-run
+    if not args.dev_run and not msaction.send_request(study_names[0], "study_exists"):
+        for study in study_names:
+            print('Requesting creation of', study)
```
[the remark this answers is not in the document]
jlchang: Looked for all the instances of concatenation that I could find to update to f-strings. Will try consistently use f-strings in future.

PR #136, opened 2020-08-19
Fix query to detect duplicate cells across expression matrices within a study (SCP-2674)

jlchang:
This PR fixes the query for obtaining the known cells in expression matrices within a study. In the current code, if a dense matrix is parsed after a metadata file, it will always fail validation because the query used to get the arrays of cells from other matrix files in that study contains the "all cells" arrays from metadata files.

This satisfies SCP-2674.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,7 +5,7 @@ with open('README.md', 'r') as fh:
 
 setup(
     name='scp-ingest-pipeline',
-    version='1.5.4',
+    version='1.5.6',
     description='ETL pipeline for single-cell RNA-seq data',
     long_description=long_description,
     long_description_content_type='text/markdown',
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -5,7 +5,7 @@
 
 setup(
     name='scp-ingest-pipeline',
-    version='1.5.4',
+    version='1.5.6',
```
eweitz, line 8: Has there been a change to the metadata convention or convention validation logic since the last PyPI release?  If not, this change seems OK but not strictly necessary, as only a new Docker image (via GitHub release) and not a PyPI release is needed.

jlchang: You're right. I didn't think through making this particular change. I don't plan on a PyPI release with this PR, just an ingest pipeline release for a new docker image.

PR #145, opened 2020-09-08
* SCP-2611 Correct gene models for dense matrices and account for single column feature files

knapii-developments:
There was a regression for parsing dense matrices where gene models were only created if the gene had expression data. The PR addresses this issue by creating gene models regardless of the presence of expression data. 

There was also a second regression. The feature file in an MTX bundle can have 1 column. As stated by [10X genomics](https://support.10xgenomics.com/single-cell-gene-expression/software/pipelines/latest/output/matrices), we now set the gene name to the gene ID in the event 1 column is present.


eweitz (approved):

devonbush (changes requested):
Code looks good.  Please add a test case that confirms gene models get created for genes with no non-zero expression scores (i.e. a test case that would fail without this fix)

knapii-developments:
> Code looks good. Please add a test case that confirms gene models get created for genes with no non-zero expression scores (i.e. a test case that would fail without this fix)

b197446

ingest/expression_files/mtx.py
```diff
@@ -190,11 +190,21 @@ def transform(self):
...
-                    prev_gene_id, prev_gene = self.genes[prev_idx - 1].split("\t")
-                    # Ff the previous gene exists, load its models
+                    features = self.genes[prev_idx - 1].split("\t")
+                    prev_gene_id = features[0]
+                    try:
+                        prev_gene = features[1]
+                    # If no gene_name field is present, hence the index error,
+                    # gene name is equivalent to gene ID
+                    except IndexError:
+                        prev_gene = prev_gene_id
```
devonbush, line 206: I think this will be more legible written as
```suggestion
                    if len(features) >= 2:
                        prev_gene = features[1]
                    # If no gene_name field is present,
                    # gene name is equivalent to gene ID
                    else:
                        prev_gene = prev_gene_id
```

devonbush (changes requested):
Needs a test to cover the single-column gene file case.  And one minor code format suggestion

devonbush (approved):

eweitz (approved):

ingest/expression_files/dense_ingestor.py
```diff
@@ -249,27 +249,19 @@ def transform(self):
...
                 data_arrays,
                 num_processed,
-                True,
+                False,
+            )
+        # Load any remaining models. This is necessary because the amount of
+        # models maybe less than the batch size.
+        if len(gene_models) > 0:
+            self.create_models(
+                [], [], None, None, gene_models, data_arrays, num_processed, True
```
jlchang, line 266: Thanks for taking the time to explain this to me, Eno!

jlchang (approved):

PR #148, opened 2020-09-23
Return metadata convention validation error messages to user (SCP-2739)

jlchang:
Users of the metadata convention reported failed validation Notifier emails that had stack traces with "unexpected exit status 65 was not ignored" and no information on how to fix their metadata files.

This simple PR changes the metadata convention from using the dev_logger to using the user_logger.

The investigation for this ticket uncovered two issues:
1. ingest_pipeline is treating numeric array-based values as invalid non-numeric data and rejecting the file as format  invalid. Created SCP-2756 to deal address this problem. Tests for array-based validation did not reveal this problem because the tests were not using the newly abstracted file format checking functions.
2. running ingest_pipeline.py produces results different from validate_metadata.py because format validation has been abstracted to a higher level and ingest_pipeline catches format errors that occur upstream of validate_metadata activity. Fixing this issue will be much more straightforward when the issue above has been resolved and the metadata validation tests can use the abstracted file format checks.

### DIFF
--- a/ingest/validation/validate_metadata.py
+++ b/ingest/validation/validate_metadata.py
@@ -897,17 +897,15 @@ def collect_jsonschema_errors(metadata, convention, bq_json=None):
         return False
 
 
-def record_issue(errfile, warnfile, issue_type, msg):
+def record_issue(issue_type, msg):
     """Print issue to console with coloring and
-    writes issues to appropriate issue file
+    send errors to logger
     """
 
     if issue_type == "error":
-        errfile.write(msg + "\n")
-        dev_logger.error(msg)
+        user_logger.error(msg)
         color = Fore.RED
     elif issue_type == "warn":
-        warnfile.write(msg + "\n")
         color = Fore.YELLOW
     else:
         color = ""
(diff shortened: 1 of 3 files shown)


eweitz (approved):

ingest/validation/validate_metadata.py
```diff
@@ -897,17 +897,15 @@ def collect_jsonschema_errors(metadata, convention, bq_json=None):
...
-def record_issue(errfile, warnfile, issue_type, msg):
+def record_issue(issue_type, msg):
     """Print issue to console with coloring and
-    writes issues to appropriate issue file
+    send errors to logger
     """
 
     if issue_type == "error":
-        errfile.write(msg + "\n")
-        dev_logger.error(msg)
```
bistline, line 907: Is there any reason why we wouldn't want to log this in the dev logger?

bistline (approved):
Looks good, just a question about additional logging.

ingest/validation/validate_metadata.py
jlchang: Sending an error to the user_logger will automatically also send it to the dev logger so we're covered.

knapii-developments (approved):

PR #149, opened 2020-09-29
Updating mtx parse to add gene model for unexpressed genes

devonbush:
This also starts the process of grouping the test data into categories to match the mock_data directories, which will hopefully make test data easier to find and manage.


jlchang (approved):
Thanks for re-organizing the test files!

bistline (approved):
Looks good, although it's a little unclear to me if we have test coverage to validate that we're getting all the gene models.

devonbush:
> Looks good, although it's a little unclear to me if we have test coverage to validate that we're getting all the gene models.

Yes, sorry that the test data change probably got muddled in all the file moving.  The AB_toy example has a gene with no expression data values, and we confirm that the count and content of the models sent to Mongo are what we expect.

PR #153, opened 2020-10-15
address invalid baseURI from EBI OLS for NCBITaxon (SCP-2820)

jlchang:
Metadata validation for NCBITaxon terms began failing on 10/14. Our term validation requests were requesting validation for NCBITAXON_XXXX terms which fail. The requests are formed using the baseURI supplied by EBI OLS. When converted to a baseURI with NCBITaxon_ instead of NCBITAXON, the validation requests succeed.

This PR adds a temporary workaround to correct the invalid baseURI we're obtaining from EBI OLS.

This resolves the bug described in SCP-2820.

### DIFF
--- a/ingest/expression_files/mtx.py
+++ b/ingest/expression_files/mtx.py
@@ -217,7 +217,6 @@ class MTXIngestor(GeneExpression):
             exp_cells.append(exp_cell)
             exp_scores.append(exp_score)
 
-
         # create gene entries for genes with no positive expression values
         for idx, gene in enumerate(self.genes):
             if not visited_expression_indices.get(idx + 1):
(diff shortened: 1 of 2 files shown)


knapii-developments (approved):
Thanks for this quick turnaround Jean!

bistline (approved):

eweitz (approved):
Any idea why changing case makes this work?

jlchang:
> Any idea why changing case makes this work?
Our input test metadata files use "NCBITaxon_9606", we parse out the ontology and the term #. I wondered where the all caps ontology name was coming from. We use the ontology name and get the baseURI from EBI OLS. They're giving us a baseURI that has all caps but it isn't valid for ontology term lookups.

My suspicion is that "NCBITaxon" is the valid ontology id for the baseURI and EBI OLS started supplying the wrong base URI to us. When querying the MONDO ontology, only "MONDO" in the baseURI is valid, "Mondo" results in the same "Resource not found" error that "NCBITAXON" generates.

PR #155, opened 2020-10-19
Update metadata convention: biosample_type optional (SCP-2806)

jlchang:
To reduce barriers to metadata convention adoption, biosample_type now optional in metadata convention.

This satisfies SCP-2806.


eweitz (approved):
Nice!  I'm glad the convention will be easier for users.

It's a bit subjective, but I could also see construing this as an improvement (not a bug fix) and bumping the convention's minor version.

eweitz:
https://github.com/broadinstitute/single_cell_portal/wiki/Required-Metadata will need an update when this goes live.

bistline (approved):

PR #158, opened 2020-10-20
* SCP-2779 Ingest raw counts

knapii-developments:
Updates ingest pipeline to support parsing of the new expression matrix type, raw count. As a consequence, we need to update the way cell uniqueness is validated. Now, cells cannot be repeated between raw counts in a single study. However, we allow repeated cells between non raw count matrices.

Check cell uniqueness raw count


ingest/expression_files/expression_files.py
```diff
@@ -138,6 +156,59 @@ def check_unique_cells(cell_names: List, study_id, client):
...
+                    # Builds study ids fields for query
+                    return generate_filters(raw_count_study_file_ids)
+                else:
+                    return None
+        else:
+            return generate_filters(raw_count_study_file_ids)
+
+    @staticmethod
+    def get_raw_count_study_file_ids(client, study_id) -> List[Dict]:
+        """Returns raw count study file ids that are apart of study"""
```
jlchang, line 194:
```suggestion
        """Returns raw count study file ids that are a part of study"""
```

jlchang (approved):
Would love a walkthrough of how the tests are set up!

ingest/expression_files/expression_files.py
```diff
@@ -138,6 +156,59 @@ def check_unique_cells(cell_names: List, study_id, client):
...
+
+        COLLECTION_NAME = "study_files"
+
+        raw_ids_query = {
+            "$and": [
+                {"study_id": study_id},
+                {"file_type": "Expression Matrix"},
```
eweitz, line 201: Blocker: this entails only raw dense matrices are supported.  We also need to support raw MTX matrices.

tests/test_expression_files.py
```diff
@@ -89,33 +89,123 @@ def mock_expression_load(self, *args):
...
+            return_value=None,
+        ):
+            header = ["GENE", "foo", "foo2"]
+            self.assertTrue(
+                GeneExpression.check_unique_cells(
+                    header, ObjectId(), ObjectId(), client_mock
+                )
+            )
+
+        """Following tests are for matrices that are raw counts"""
```
eweitz, line 138: Conventional single-line comment would be better here:

```suggestion
        # Following tests are for matrices that are raw counts
```

tests/test_expression_files.py
```diff
@@ -89,33 +89,123 @@ def mock_expression_load(self, *args):
...
         client_mock["data_arrays"].find.return_value = [
             {"values": ["foo3", "foo4", "foo5"]},
             {"values": ["foo6", "foo7", "foo8"]},
         ]
-        with self.assertRaises(ValueError) as cm:
-            GeneExpression.check_unique_cells(header, ObjectId(), client_mock)
-        self.assertTrue("contains 1 cells" in str(cm.exception))
-        self.assertTrue("foo3" in str(cm.exception))
 
+        """Following tests are for matrices that aren't raw counts"""
```
eweitz, line 98: Conventional single-line comment would be better here:

```suggestion
        # Following tests are for matrices that aren't raw counts
```

eweitz (changes requested):
Code looks good, but it seems there's a functional problem with raw MTX counts.

Also, a non-blocking idea maybe for later:

Would it be worth validating that raw count values have no significant digits after the decimal point -- i.e. that they are counts?  This would e.g. catch user error in uploading a visualization matrix as a count matrix.

ingest/expression_files/expression_files.py
```diff
@@ -112,12 +115,27 @@ def check_unique_cells(cell_names: List, study_id, client):
             "$nor": [{"name": "All Cells"}],
         }
         # Returned fields from query results
-        field_names = {"values": 1, "_id": 0}
+        FIELD_NAMES = {"values": 1, "_id": 0}
+        COLLECTION_NAME = "data_arrays"
+
```
devonbush, line 120: I don't see the part of this method that handles checking unique values in non-raw-counts files.  Let's touch base today so you can talk me through the approach -- I'm probably missing something

devonbush (changes requested):
Let's talk, I'm not sure I understand the logic flow

ingest/expression_files/expression_files.py
```diff
@@ -93,17 +93,20 @@ def create_gene_model(
...
          Parameters:
             cell_names (List[str]): List of cell names in matrix
             study_id (ObjectId): The study id the cell names belong to
+            study_file_id (ObjectId): The file id the cell names belong to
             client: MongoDB client
         """
-        COLLECTION_NAME = "data_arrays"
-        query = {
+        QUERY = {
```
bistline, line 109: I think we have a disconnect in the logic flow here.  Previously, we only had one "bucket" of cells to check for repeats when validating that cells were unique (all expression matrices).  Now, we have two separate buckets: normalized "visualization" matrices, and raw counts matrices.  The logic for validation is the same - the only thing that we're changing is which "bucket" we use to validate against.

To achieve this, I think all that needs to happen is to append an `$in` operator for `study_file_id` to the `QUERY` object.  The contents of that array will be all expression matrices of the same "type" as the one being ingested - either raw counts, or not.  We can easily do this by getting _all_ expression matrices for the study (minus the one being ingested), and filtering out the ones that aren't of the same "type" as the matrix being ingested.

Example (this can probably be refined, but will show the logic):
```
ingest_file =  client['study_files'].find({"id": study_file_id})
# we need to check for the 'expression_file_info' object first
is_raw_counts = ingest_file['expression_file_info'] is not None && 
    ingest_file['expression_file_info']['is_raw_counts']

study_file_ids = []
study_files = client['study_files'].find({
    "study_id": study_id, 
     "$in": [
        "file_type": ['Expression Matrix', 'MM Coordinate Matrix']
     ], 
    "$not": {"id": study_file_id}
)

# grab the file id of all matrix files that are of the same type
for file in study_files:
    file_is_raw_counts = file['expression_file_info'] is not None && 
        file['expression_file_info']['is_raw_counts']
    if is_raw_count == file_is_raw_counts
        study_file_ids.append(file['id'])
        
QUERY = {
    "$and": [
        {"linear_data_type": "Study"},
        {"array_type": "cells"},
        {"study_id": study_id},
    ],
    "$nor": [{"name": "All Cells"}],
    "$in": [{"study_file_id": study_file_ids}]
}
...
```

After this, we can look for intersection as normal.

bistline (changes requested):
I don't think that the validation logic here is doing what we want (see comment).

ingest/expression_files/expression_files.py
```diff
@@ -93,17 +93,20 @@ def create_gene_model(
         )
 
     @staticmethod
-    def check_unique_cells(cell_names: List, study_id, client):
+    def check_unique_cells(cell_names: List, study_id, study_file_id, client):
         """Checks cell names against database to confirm matrix contains unique
             cell names.
 
+        Method acts as 'command center' for deciding what methods need to be called to check for cell uniquenes under
```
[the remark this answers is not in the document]
knapii-developments: cb3dd48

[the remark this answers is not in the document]
knapii-developments: cb3dd48

ingest/expression_files/expression_files.py
knapii-developments: 8a03932

ingest/expression_files/expression_files.py
knapii-developments: 0475ff4

ingest/expression_files/expression_files.py
```diff
@@ -138,6 +158,66 @@ def check_unique_cells(cell_names: List, study_id, client):
...
+        filters = []
+        for results in query_results:
+            for field in field_values:
+                field_key = query_mappings[field]
+                field_value = results[field]
+                filters.append({field_key: field_value})
+        return filters
+
+    @staticmethod
+    def get_study_file_ids_by_type(
```
bistline, line 196: Since we're in the `ExpressionFiles` class, and this method (as written) will only ever return expression matrices: 
```suggestion
    def get_study_expression_file_ids(
```

Or something similar.  Since we're not passing `file_type` as a parameter, it seems somewhat misleading to name the method this.

bistline (changes requested):
Very nice refactoring!  Just some corrections on queries to make sure that we are scoping things correctly.

ingest/expression_files/expression_files.py
```diff
@@ -138,6 +158,66 @@ def check_unique_cells(cell_names: List, study_id, client):
             raise ValueError(error_string)
         return True
 
+    @staticmethod
+    def has_expression_file_info_doc(client, study_id):
```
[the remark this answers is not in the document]
knapii-developments: 461741a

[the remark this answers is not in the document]
bistline: 461741a doesn't appear to be a commit - I'm not seeing the update here.

ingest/expression_files/expression_files.py
```diff
@@ -93,17 +93,25 @@ def create_gene_model(
...
-            cell names.
+    def is_raw_count(study_file_id, client):
+        COLLECTION_NAME = "study_files"
+        FIELD_NAMES = {"expression_file_info.is_raw_counts": 1, "_id": 0}
+        QUERY = {"_id": study_file_id}
+        query_results = list(client[COLLECTION_NAME].find(QUERY, FIELD_NAMES)).pop()
+        return query_results["expression_file_info"]["is_raw_counts"]
+
+    @staticmethod
+    def check_unique_cells(cell_names: List, study_id, study_file_id, client):
```
devonbush, line 104: I think this method as a whole would be easier to read and test if you removed all mongo query logic from it, so this method just reads as a list of logical steps, and the mongo interfacing is delegated out.  E.g. 

```
check_unique_cells(cell_names, study_id, study_file_id, client):
    study_file = get_study_file_by_id(study_file_id)
    is_raw_counts = is_raw_counts_file(study_file)
    cells_of_same_type = get_study_cells_by_type(study_id, is_raw_counts)
    <<compare cells_of_same_type and cell_names for dupes>>

```

ingest/expression_files/expression_files.py
```diff
@@ -138,6 +158,66 @@ def check_unique_cells(cell_names: List, study_id, client):
...
+
+        :parameter
+            query_results (List[Dict]) : Results from any query
+            field_values (List) : Keys that are taken from query_results
+            query_mappings : Map query_results fields to model attributes
+        Example:
+               query_results = [{_id: 1234}, {food:abc}, {barbara: streisand}]
+               field_values = [id, food]
+               query_mappings = {id: study_file_id, barbara: lastname}
+
```
devonbush, line 184: I don't understand what happened to the 'food' value in this example

devonbush (changes requested):

ingest/expression_files/expression_files.py
```diff
@@ -112,12 +120,24 @@ def check_unique_cells(cell_names: List, study_id, client):
             "$nor": [{"name": "All Cells"}],
         }
         # Returned fields from query results
-        field_names = {"values": 1, "_id": 0}
+        FIELD_NAMES = {"values": 1, "_id": 0}
+        COLLECTION_NAME = "data_arrays"
+
+        study_files_ids = GeneExpression.get_study_file_ids_by_type(
```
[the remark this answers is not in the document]
knapii-developments: See [comment](https://github.com/broadinstitute/scp-ingest-pipeline/pull/158/files/06057fbedd1d0ad3cd7c1152e4bb96deb815d6d1#r511051170). I'm indifferent about what the function is named. If this seems clearer let me know!

ingest/expression_files/expression_files.py
knapii-developments: I'm not convinced that it would be easier to test. The same methods still need to be mocked. The query also needs to be validated for every test case until we have a mongo emulator. 

However, I did abstract out the mongo query logic and the code does look cleaner! Thanks for making the suggestions. 

99313db

ingest/expression_files/expression_files.py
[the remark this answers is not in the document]
knapii-developments: I just pushed :D

ingest/expression_files/expression_files.py
knapii-developments: This function was removed

ingest/expression_files/expression_files.py
knapii-developments: [99313db](https://github.com/broadinstitute/scp-ingest-pipeline/blob/097b087cf1298b0d2143f92d7b748b32232b92c0/ingest/expression_files/expression_files.py#L187)

ingest/expression_files/expression_files.py
```diff
@@ -93,17 +93,25 @@ def create_gene_model(
         )
 
     @staticmethod
-    def check_unique_cells(cell_names: List, study_id, client):
-        """Checks cell names against database to confirm matrix contains unique
-            cell names.
+    def is_raw_count(study_file_id, client):
```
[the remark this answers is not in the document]
knapii-developments: I'm not sure what you mean here. This has already been done I believe....

ingest/expression_files/expression_files.py
```diff
@@ -93,17 +93,25 @@ def create_gene_model(
...
+        COLLECTION_NAME = "study_files"
+        FIELD_NAMES = {"expression_file_info.is_raw_counts": 1, "_id": 0}
+        QUERY = {"$and": [{"_id": study_file_id}, {"file_type": "Expression Matrix"}]}
+        query_results = list(client[COLLECTION_NAME].find(QUERY, FIELD_NAMES)).pop()
+        return query_results["expression_file_info"]["is_raw_counts"]
```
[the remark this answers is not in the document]
knapii-developments: Good point! I added a check [here](https://github.com/broadinstitute/scp-ingest-pipeline/blob/097b087cf1298b0d2143f92d7b748b32232b92c0/ingest/expression_files/expression_files.py#L102) and the tests that follow the logic in this commit 602b2e1

bistline (approved):

ingest/expression_files/expression_files.py
eweitz: [SCP-2000](https://broadworkbench.atlassian.net/browse/SCP-2000) tracks integrating a MongoDB emulator.

ingest/expression_files/expression_files.py
```diff
@@ -93,31 +93,63 @@ def create_gene_model(
...
+            study_id (ObjectId): The study id the cell names belong to
+            study_file_id (ObjectId): The file id the cell names belong to
+            client: MongoDB client
+        """
+
+        additional_query_kwargs = {}
+        study_files_ids = GeneExpression.get_study_expression_file_ids(
+            study_id, study_file_id, client
+        )
+
```
devonbush, line 141: I would pull everything from lines 142-160 into a helper method "get_cell_names_from_study_file_ids" or something like that.  The method can just return an empty array if the study_file_id array is empty

devonbush (changes requested):
Functionally it's looking good, just some changes needed for simplicity & readability

eweitz (approved):

ingest/expression_files/expression_files.py
```diff
@@ -93,31 +93,63 @@ def create_gene_model(
...
-    def check_unique_cells(cell_names: List, study_id, client):
-        """Checks cell names against database to confirm matrix contains unique
-            cell names.
+    def is_raw_count(study_id, study_file_id, client):
+        COLLECTION_NAME = "study_files"
+        FIELD_NAMES = {"expression_file_info.is_raw_counts": 1, "_id": 0}
+        QUERY = {"_id": study_file_id}
+        # If study files does not have document expression_file_info
+        # field, "is_raw_counts", will not exist.
+        if GeneExpression.has_expression_file_info_doc(study_id, study_file_id, client):
```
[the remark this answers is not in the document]
knapii-developments: 6d8f85d

ingest/expression_files/expression_files.py
knapii-developments: 7352d5c

ingest/expression_files/expression_files.py
```diff
@@ -93,31 +93,63 @@ def create_gene_model(
         )
 
     @staticmethod
-    def check_unique_cells(cell_names: List, study_id, client):
-        """Checks cell names against database to confirm matrix contains unique
-            cell names.
+    def is_raw_count(study_id, study_file_id, client):
```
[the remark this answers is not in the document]
knapii-developments: 88d8a72

ingest/expression_files/expression_files.py
```diff
@@ -138,6 +170,49 @@ def check_unique_cells(cell_names: List, study_id, client):
...
+            "study_file_id": study_file_id,
+            "expression_file_info": {"$exists": True},
+        }
+        query_results = list(client[COLLECTION_NAME].find(QUERY))
+        if query_results:
+            return True
+        return False
+
+    @staticmethod
+    def get_study_expression_file_ids(
```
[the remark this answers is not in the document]
knapii-developments: I named it this way because of this [comment](https://github.com/broadinstitute/scp-ingest-pipeline/pull/158#discussion_r511051170). If you'd still like it changed I'll be happy to switch it up!

PR #159, opened 2020-10-26
Update scientific python package dependency versions SCP-2802

jlchang:
Pinning pandas, scipy, and numpy to more recent releases.
With upgrade of Pandas to 1.X, duplicate headers are no longer allowed. Invoking SCP's validate_unique_header earlier in Annotations.py (and shifting annotation file format tests out of  metadata validation to annotation tests) sidesteps issues with the Pandas behavior change.

This satisfies SCP-2802.


ingest/annotations.py
```diff
@@ -86,7 +86,15 @@ def preprocess(self):
         # Uppercase NAME and TYPE
         self.headers[0] = self.headers[0].upper()
         self.annot_types[0] = self.annot_types[0].upper()
-        self.create_data_frame()
+        if self.validate_unique_header():
+            self.create_data_frame()
+        else:
+            msg = (
+                "Unable to parse file - Duplicated annotation header names are not allowed. \n"
```
eweitz, line 93: "Duplicated" isn't wrong, but "duplicate" seems more idiomatic.  [This thread](https://english.stackexchange.com/q/67463) helped refine my thinking.  The latter is also a bit more internally consistent with the code.

```suggestion
                "Unable to parse file - Duplicate annotation header names are not allowed. \n"
```

ingest/annotations.py
```diff
@@ -175,13 +183,13 @@ def validate_unique_header(self):
             for x in self.headers:
                 if x in seen_headers or seen_headers.add(x):
                     duplicate_headers.add(x)
-            msg = (
-                f"Duplicated metadata header names are not allowed: {duplicate_headers}"
-            )
+            msg = f"Duplicated header names are not allowed: {duplicate_headers}"
```
eweitz, line 186:
```suggestion
            msg = f"Duplicate header names are not allowed: {duplicate_headers}"
```

eweitz (approved):
Looks good!  I noted a nit.

ingest/annotations.py
jlchang: I agree "Duplicate" is more appropriate and more consistent. Thanks for the suggestion!

bistline (approved):

PR #161, opened 2020-10-28
* SCP-2761 Measure the number of unsorted/sorted mtx uploads via Mixpanel

knapii-developments:
This PR allows tracking of unsorted matrices. Mixpanel will log if an MTX file is unsorted and if it is, how long it takes to sort the file. 
Screen Shot 2020-11-12 at 3 00 11 PM


ingest/expression_files/mtx.py
```diff
@@ -202,6 +206,7 @@ def get_features(feature_row: str):
         return gene_id, gene_name
 
     @staticmethod
+    @custom_metric(config.get_metric_properties, "sortPerfTime", {"sorted": False})
```
eweitz, line 209: Please use `perfTime:sort` (generally `<perfTime:<foo>`), [like `single_cell_portal_core`](https://github.com/broadinstitute/single_cell_portal_core/pull/716):

```suggestion
    @custom_metric(config.get_metric_properties, "perfTime:sort", {"sorted": False})
```

eweitz (approved):
I approved, assuming the suggested event name change will be committed.

jlchang (approved):

ingest/expression_files/mtx.py
knapii-developments: 317fded

PR #163, opened 2020-11-03
* SCP-2796 Integrate ingest pipeline and Mixpanel

knapii-developments:
This PR adds logging to Mixpanel via Bard. Particularly, properties study accession, file name, type, and size, and ingest time for all ingested files.  

I'm looking for feedback on this implementation before writing tests. This will address build failures.

As mentioned in this [comment](https://github.com/broadinstitute/scp-ingest-pipeline/pull/163#discussion_r516951448), perfTime encompasses how long it takes to complete an entire ingest run. I refer to this as the parent event. A parent event encompasses all child events, which are decorated functions, and their properties. All child events will have their own perfTimes that are represented in the parent event. 
Definition of mixpanel event and properties

 Here is what the outcome is in Mixpanel:
Screen Shot 2020-11-05 at 11 32 05 AM


ingest/monitoring/metrics_service.py
```diff
@@ -0,0 +1,33 @@
...
+# Mixpanel is a business analytics service for tracking user interactions.
+# Bard is a DSP service that mediates writes to Mixpanel.
+#
+
+import json
+import requests
+
+
+class MetricsService:
+    BARD_HOST_URL = "https://terra-bard-dev.appspot.com/api/event"
```
bistline, line 12: We're going to need a way for this to work in all environments, so hard-coding the URL like this won't work.

ingest/monitoring/metrics_service.py
```diff
@@ -0,0 +1,33 @@
...
+    user_id = "2f30ec50-a04d-4d43-8fd1-b136a2045079"
```
bistline, line 13: Same thing - why is the `user_id` hard-coded?

bistline (changes requested):
Values in `metrics_service.py` being hard-coded won't work for actual usage.

ingest/monitoring/metrics_service.py
knapii-developments: From my understanding, it doesn't matter. @devonbush can provide more context.

ingest/settings.py
```diff
@@ -0,0 +1,64 @@
...
+    FILE_NAME = None
+
+    def __init__(self, study_file_id):
+        StudyFile.update(study_file_id)
+
+    @classmethod
+    def update(cls, study_file_id):
+        cls.STUDY_FILE_ID = study_file_id
+        query = MONGO_CONNECTION._client["study_files"].find(
+            {"_id": ObjectId("5e3dd9bcfa6429263aab24ba")},
```
eweitz, line 58: This looks like an unintentionally-hardcoded study file ID.

ingest/settings.py
```diff
@@ -0,0 +1,64 @@
...
+            {"upload_file_size": 1, "file_type": 1, "name": 1, "_id": 0},
+        )
+        query_results = list(query).pop()
+        cls.FILE_TYPE = query_results["file_type"]
+        cls.FILE_SIZE = query_results["upload_file_size"]
+        cls.FILE_NAME = query_results["name"]
```
eweitz, line 64: Like `study_accession`, passing these in as arguments from Rails to `ingest_pipeline.py` seems better than querying them from here.

ingest/settings.py
```diff
@@ -0,0 +1,64 @@
...
+        cls.FILE_TYPE = file_type
+
+    @classmethod
+    def update_study_accession(cls, study_id):
+        study_accession = list(
+            MONGO_CONNECTION._client["study_accessions"].find(
+                {"study_id": ObjectId(study_id)}, {"accession": 1, "_id": 0}
+            )
+        ).pop()
+        cls.STUDY_ACCESSION = study_accession["accession"]
```
eweitz, line 42: It seems faster, simpler, and more robust to pass `study_accession` as an input argument to `ingest_pipeline.py` from Rails.

ingest/settings.py
```diff
@@ -0,0 +1,64 @@
...
+
+class Study:
+    STUDY_ACCESSION: str = None
+
+    def __init__(self, study_id):
+        Study.update_study_accession(study_id)
+
+    @classmethod
+    def update_file_type(cls, file_type):
+        cls.FILE_TYPE = file_type
```
eweitz, line 33: This seems unused.

```suggestion
```

ingest/monitoring/metrics_service.py
eweitz: I suggest passing `BARD_HOST_URL` as an input parameter to `ingest_pipeline.py`.  This would let us keep one central list of environment-to-Bard-host mappings, [as defined in `single_cell_portal_core`](https://github.com/broadinstitute/single_cell_portal_core/search?q=bard_host_url&type=code).

Not using this suggested approach caused a bug in another SCP Bard client; see SCP-2866.

ingest/ingest_pipeline.py
```diff
@@ -293,6 +299,9 @@ def upload_metadata_to_bq(self):
                 return 1
         return 0
 
+    @custom_metric(
+        "ingest-pipeline:expression:ingest", settings.get_study, settings.get_study_file
```
eweitz, line 303: This approach lets us analyze across jobs -- e.g. median duration of the `ingest-pipeline:expression:ingest` event across many ingests.  The downside is that it's quite network-intensive (one log request per function call) given that the only props that change are `functionName` (which maps 1-to-1 with `event`) and `perfTime`.

A complementary approach to consider: we could also accumulate these events and their `perfTimes` in your new singleton object.  Then, when ingest completes (successfully or not), we could log _one event_ with an ordered list of `perfTime:<event_order>:<event_name>` props for all events in a given ingest.  

This would follow a pattern we use in `single_cell_portal_core` (see [diagram and explanation](https://github.com/broadinstitute/single_cell_portal_core/pull/716)) that we've found helpful in Mixpanel triage.  I don't think it's blocking or necessarily in scope here -- just thought I'd note it for your consideration.  It would augment your existing work here.

eweitz (changes requested):
Good progress!

See comments in previous code review (I started them between review request and PR being marked ready, and didn't notice the status change).

ingest/settings.py
devonbush: I dunno, Ingest Pipeline already has to know the intricacies of our study file/data array data models.  So I like keeping the interface between portal_core and pipeline simple -- passing in more and more args makes command-line testing of ingest pipeline more difficult

ingest/settings.py
```diff
@@ -0,0 +1,64 @@
+# File is responsible for defining globals and initializing them
+from mongo_connection import MongoConnection
+from bson.objectid import ObjectId
+
+MONGO_CONNECTION = MongoConnection()
+
+
+def init(study_id, study_file_id):
```
devonbush, line 8: Good stuff.  I did some research and realized why we were talking past each other so much on Monday.  'global' in python is actually module-scoped, which is what I want.  I didn't realize that and thought when you were saying these were global, that they were not module-scoped.  so we're on the same page -- I just didn't know the pythonese  

I believe best practice is to name variables like this with double-underscores. e.g.   __study__  to indicate that other modules are not supposed to access this variable directly, and instead go through your getters

ingest/settings.py
```diff
@@ -0,0 +1,64 @@
...
+
+def get_study_file():
+    global study_file
+    return study_file
+
+
+class Study:
+    STUDY_ACCESSION: str = None
+
+    def __init__(self, study_id):
```
devonbush, line 28: I think this will be cleaner if __init__ calls a method like 'fetch_study', that stores the study document, and then you just make getters get_accession, and get_file_type that read from the study object

ingest/settings.py
```diff
@@ -0,0 +1,64 @@
...
+        cls.STUDY_ACCESSION = study_accession["accession"]
+
+
+class StudyFile:
+    FILE_TYPE: str = None
+    FILE_SIZE: int = None
+    STUDY_FILE_ID = None
+    FILE_NAME = None
+
+    def __init__(self, study_file_id):
```
devonbush, line 51: similar to above, 'fetch_study' is probably a more descriptive name for .update.  Then it can just store the document in a variable STUDY_FILE, and you can make getters for the things you need

ingest/settings.py
```diff
@@ -0,0 +1,64 @@
...
+        StudyFile.update(study_file_id)
+
+    @classmethod
+    def update(cls, study_file_id):
+        cls.STUDY_FILE_ID = study_file_id
+        query = MONGO_CONNECTION._client["study_files"].find(
+            {"_id": ObjectId("5e3dd9bcfa6429263aab24ba")},
+            {"upload_file_size": 1, "file_type": 1, "name": 1, "_id": 0},
```
devonbush, line 59: I don't think you want to have these criteria above -- you should just be getting the whole document by id. StudyFiles are small objects.

devonbush (dismissed):
I like where this is at -- I know I was skeptical on Monday, but looking at your code (and doing more python research), I realize that we were actually on the same page.  We were just using different words.  You've created a module-scoped logger, and a module-scoped config object, with appropriate methods to expose access.  Which is great.  Just some refinements needed around method names, and then also Eric's blocking concerns

ingest/settings.py
eweitz: You're right that the interface is simpler this way, but that also makes the pipeline more complex to test and maintain.  

Because we can't test database querying in Ingest Pipeline itself (due to lacking a MongoDB emulator, SCP-2000), the current approach means bugs in querying won't get caught until portal_core integration tests, at best.  Even after SCP-2000, this way entails more involved mocks to get the code covered.

By contrast, passing these as arguments lets us test them like other arguments, and means less code overall.

I defer to Eno.  This approach is reasonable, if not my ideal, and at least improves the case for a MongoDB emulator!

ingest/ingest_pipeline.py
```diff
@@ -50,6 +50,10 @@
     # Used when importing internally and in tests
     from ingest_files import IngestFiles
 
+    # For Mixpanel logging
+    import settings
+    from montoring.mixpanel_log import custom_metric
```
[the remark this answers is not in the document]
knapii-developments: Interesting that flake8 didn't catch this....

ingest/settings.py
knapii-developments: 1e1779a

ingest/ingest_pipeline.py
```diff
@@ -310,13 +318,14 @@ def ingest_expression(self) -> int:
...
                     **self.kwargs,
-                )  # Dense receiver
+                )
             self.expression_ingestor.execute_ingest()
         except Exception as e:
             log_exception(IngestPipeline.dev_logger, IngestPipeline.user_logger, e)
             return 1
         return 0
 
+    @custom_metric(config.get_metric_properties)
```
bistline, line 328: Very nice!

bistline (approved):
Very cool - and thanks for explaining the `testing_guard` decorator.

ingest/monitoring/mixpanel_log.py
```diff
@@ -0,0 +1,74 @@
...
+        # Calculate performance time
+        duration = end_time - self.start_time
+        self.props[self.perf_time_name] = round(duration, 3)
+        # Add current functions properties to base model
+        self.metric_properties.update(self.props)
+        # Add child events to list of child events captured
+        if self.child_event_name:
+            add_child_event(self.child_event_name)
+
+
```
eweitz, line 42:
```suggestion
# @testing_guard is applied here so Mixpanel events are not logged during tests.
```

tests/conftest.py
```diff
@@ -0,0 +1,13 @@
+import pytest
+
+
+@pytest.fixture(autouse=True)
+def set_testing(monkeypatch):
+    """Sets the TESTING environment variable."""
+    monkeypatch.setenv("TESTING", "")
```
eweitz, line 7: Neat!

eweitz (approved):
Looks good!  I suggested a few maintainability and style refinements.  

Thanks especially for the diagram.  It conveys how this logging uses similar patterns as [logging in `single_cell_portal_core`](https://github.com/broadinstitute/single_cell_portal_core/pull/716) -- which would be much less apparent without your visual explanation.

ingest/monitoring/mixpanel_log.py
```diff
@@ -0,0 +1,74 @@
...
+        function_name,
+        perf_time_name,
+        metric_properties: Dict,
+        props: Dict = None,
+        child_event_name: str = None,
+    ):
+        props["functionName"] = function_name
+        self.child_event_name = child_event_name
+        self.props = props
+        self.perf_time_name = perf_time_name if perf_time_name else "perfTime"
```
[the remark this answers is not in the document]
knapii-developments: This ia little hard to explain. But basically the default perfTime name is perfTime. For child events, the perfTime would be named after the child event. For example, the [perfTime for sorted MTX files](https://github.com/broadinstitute/scp-ingest-pipeline/blob/39e1715c44c68f4f81c553ec593615b1ce6864dc/ingest/expression_files/mtx.py#L209) is named  "isSortedPerfTime". I could add a comment or maybe do a slight refactor such as:

```
default_perfTime_name = "perfTime"
perf_time_name = perf_time_name if perf_time_name else default_perfTime_name
```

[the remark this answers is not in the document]
eweitz: Your original logic seems clear, I just think it'd just be clearer and less code to use default parameter values in the method signature.

Also (more importantly), please use the form `perfTime:<foo>` instead of `<foo>PerfTime`  (e.g. `perfTime:isSorted` instead of `isSortedPerfTime`) for consistency with such logging in `single_cell_portal_core`.

ingest/settings.py
knapii-developments: I believe we paired on this. I refactored this setup. I isolated [study](https://github.com/broadinstitute/scp-ingest-pipeline/blob/626b12150350eec3a08e60b76202418e69dc701f/ingest/config.py#L58) and[ study files](https://github.com/broadinstitute/scp-ingest-pipeline/blob/626b12150350eec3a08e60b76202418e69dc701f/ingest/config.py#L89) to have their own classes where each class has access to their respective documents. I used the ```@property``` decorator so that they have getter and setter methods.

ingest/settings.py
knapii-developments: See [comment](https://github.com/broadinstitute/scp-ingest-pipeline/pull/163#discussion_r522325975)

ingest/settings.py
knapii-developments: See [comment](https://github.com/broadinstitute/scp-ingest-pipeline/pull/163#discussion_r522325975)

ingest/monitoring/mixpanel_log.py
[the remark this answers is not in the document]
knapii-developments: Just for my understanding perfTime is a number. So the perfTimes of child events are named like <child event name>PerfTime. A single ingest run captures all perfTimes of decorated functions. Please see the example [here](https://github.com/broadinstitute/scp-ingest-pipeline/pull/161#issue-511678735). The field value "perfTime" captures the total amount it took to parse and store.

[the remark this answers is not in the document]
knapii-developments: 0f77b31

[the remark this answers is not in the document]
eweitz: Yup, `perfTime` is the entire time as a number -- including the sum of all children.

But for consistency with other logging, those child perfTimes property names should take the form `perfTime:<child event name>` (e.g. `"perfTime:isSorted": 1234`) and not `<child event name>PerfTime`.

ingest/settings.py
knapii-developments: See [comment](https://github.com/broadinstitute/scp-ingest-pipeline/pull/163#discussion_r522325975)

ingest/ingest_pipeline.py
knapii-developments: Ohhhhh ok. I'll make this update in the other PR. I appreciate your attention to detail on this!

PR #167, opened 2020-11-13
* SCP-2891 Add extra validations for gzipped files

knapii-developments:
This PR adds extra logic for determining if a file is gzipped or not.


eweitz (approved):
I approve.  We'll make an exception for the two-review policy, given:
* The SCP team is short-staffed today -- Eno and I are the only implementers available; and
* This fixes an issue that actively blocks at least two users

Feel free to merge when tests pass, Eno.

PR #170, opened 2020-11-13
* SCP-2796 Fix URL for Bard

knapii-developments:
An incorrect URL from ingest pipeline was causing [build errors](https://travis-ci.com/github/broadinstitute/single_cell_portal_core/builds/201355050#L2935) when releasing scp_core. This PR addresses the issue by appending the endpoint to the base URL.


eweitz (approved):
I approve.  We'll make an exception to SCP's two-review policy, given:

* SCP is short-staffed today -- Eno and I are the only implementers available; and
* A release is scheduled for Monday morning

PR #173, opened 2020-11-16
Ingest Pipeline 1.8.3

jlchang:
This PR updates Ingest Pipeline from version 1.8.2 to 1.8.3.


eweitz (approved):

bistline (approved):

ingest/expression_files/mtx.py
```diff
@@ -191,14 +191,20 @@ def get_data_start_line_number(file_handler: IO) -> int:
...
-                # First line w/o '%' is mtx dimension. So skip this line (+1)
-                i += 2
-                return i
+        for count, line in enumerate(file_handler):
+            if not line.startswith("%"):
+                try:
+                    line_values = line.strip().split()
+                    float(line_values[0])
+                    # First line w/o '%' is mtx dimension. So skip this line (+1)
+                    count += 2
```
devonbush, line 200: you shouldn't modify the count variable inside a loop like this--modifying iterators inside a loop is generally dangerous and confusing b/c it suggests that you've actually moved the file pointer.  just do `return count  + 2`

devonbush (approved):
one stylistic change, then this looks good

knapii-developments (approved):

ingest/expression_files/mtx.py
jlchang: Eno made this change in PR #175

PR #174, opened 2020-11-16
* SCP-2900 Resolve "not enough values to unpack (expected 3, got 1)"

knapii-developments:
PR addresses valid MTX files with an empty comment line after the expected header causing "not enough values to unpack (expected 3, got 1)" in production.

### DIFF
--- a/ingest/expression_files/mtx.py
+++ b/ingest/expression_files/mtx.py
@@ -125,7 +125,7 @@ class MTXIngestor(GeneExpression, IngestFiles):
             # numeric(-n). Use stable sort (--stable) so that columns are only compared by the first column.
             # Without this argument line '1 3 4' and ' 1 5 4' would be considered unsorted.
             p2 = subprocess.run(
-                ["sort", "-s", "-c", "--stable", "-n", "-k", "1,1"],
+                ["sort", "-c", "--stable", "-n", "-k", "1,1"],
                 stdin=p1.stdout,
                 capture_output=True,
             )
--- a/ingest/expression_files/mtx.py
+++ b/ingest/expression_files/mtx.py
@@ -192,14 +192,17 @@ class MTXIngestor(GeneExpression, IngestFiles):
             i (IO): Line number where data starts
         """
         for count, line in enumerate(file_handler):
-            try:
-                line_values = line.split()
-                float(line_values[0])
-                # First line w/o '%' is mtx dimension. So skip this line (+1)
-                count += 2
-                return count
-            except ValueError:
-                pass
+            if not line.startswith("%"):
+                try:
+                    line_values = line.split()
+                    float(line_values[0])
+                    # First line w/o '%' is mtx dimension. So skip this line (+1)
+                    count += 2
+                    return count
+                except ValueError:
+                    raise ValueError(
+                        "Only header, comment lines starting with '%', and numeric data allowed in MTX file."
+                    )
         raise ValueError(
             "MTX file did not contain expression data. Please check formatting and contents of file."
         )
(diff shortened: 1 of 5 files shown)


ingest/expression_files/mtx.py
```diff
@@ -193,12 +193,14 @@ def get_data_start_line_number(file_handler: IO) -> int:
         """
         i = 0
         for line in file_handler:
```
devonbush, line 195: ```for count, line in enumerate(file_handler):``` seems like a more elegant/pythonic way to get the line count like you want

devonbush (commented):
looks like a good, robust approach -- just keep going until we find something that's a number.  You'll want to add a bunch of tests though :)

knapii-developments:
> looks like a good, robust approach -- just keep going until we find something that's a number. You'll want to add a bunch of tests though :)

I am adding tests. However, anything that isn't a % or digit should throw an error because the file is not compliant with MTX format. But it's hard to test whether this was this is a developer, which this one is, or user error.

tests/data/mtx/mtx_character_before_data.mtx
```diff
@@ -0,0 +1,8 @@
+%%MatrixMarket matrix coordinate real general
+a
```
jlchang, line 2: This line is invalid for the file format and should trigger the ValueError ""MTX file did not contain expression data"

tests/test_mtx.py
```diff
@@ -93,6 +93,8 @@ def test_get_data_start_line_number(self):
         mtx_file_handler = open("data/mtx/unsorted_mtx.mtx.txt")
         self.assertEqual(3, MTXIngestor.get_data_start_line_number(mtx_file_handler))
 
+        mtx_file_handler = open("data/mtx/mtx_character_before_data.mtx")
+        self.assertEqual(5, MTXIngestor.get_data_start_line_number(mtx_file_handler))
```
jlchang: Processing mtx_character_before_data.mtx should result in a ValueError because the MTX file does not conform to the file format spec. A separate test file with extra comment lines (ie. prefixed with the % symbol) with "a" or "/" as values should be used for this particular test where we expect that the new start line is 5.

jlchang (changes requested):
The test passes in an invalid file which should fail validation. Please fix to check and ignore valid comment lines but not invalid lines in the file. I'd recommend amending the invalid file to have valid comment lines (so the existing test tests that it gets a correct new line number) and have a separate test that check for invalid file format of a non-comment line before the expected line for matrix dimension numbers.

ingest/expression_files/mtx.py
```diff
@@ -191,29 +191,34 @@ def get_data_start_line_number(file_handler: IO) -> int:
...
-                i += 1
-            else:
-                # First line w/o '%' is mtx dimension. So skip this line (+1)
-                i += 2
-                return i
+        for count, line in enumerate(file_handler):
+            if not line.startswith("%"):
+                try:
+                    line_values = line.split()
+                    float(line_values[0])
```
eweitz, line 198: To help maintainability:

```suggestion
                    float(line_values[0]) # Determines if value is numeric
```

eweitz (approved):

tests/data/mtx/bad_format_has_space.mtx
```diff
@@ -0,0 +1,12 @@
+
```
jlchang, line 1: Is this file bad because of this first empty line or the line after the comment lines? From the code, it sounds like we're checking for the first line. Is line 5 important in these tests, if not, can we remove it to be explicit about what we are testing?

knapii-developments: It keeps checking until a valid or invalid character is reached. If neither happens it's an empty file.

jlchang (approved):
Thanks for making the changes to the tests!

ingest/expression_files/mtx.py
knapii-developments: ad87eb7

PR #177, opened 2020-12-02
Document reasoning for Mouse Brain Atlas ID formatting for SCP ingest (SCP-2918)

jlchang:
Mouse Brain Atlas uses numeric IDs for their ontology terms. To prevent name collision, it seemed prudent to uniquefy the ID using the syntax used by other ontologies - a full name comprised of an ontology shortname, separated by underscore or colon, followed by an id. We chose "MBA" to be the ontology shortname for Mouse Brain Atlas (at the time of choosing, it was not in use as a shortname for any other ontology). The longest ID in MBA is 9 digits - we requested that ids shorter than 9 digits be padded with leading zeros in the fashion of MONDO, PATO and other ontologies.

Our U19 collaborators suggested we document the reasoning above for future reference.

### DIFF
--- a/ingest/validation/validate_metadata.py
+++ b/ingest/validation/validate_metadata.py
@@ -179,6 +179,13 @@ class OntologyRetriever:
     def retrieve_ontology_term_label_remote(self, term, property_name, ontology_urls):
         """Look up official ontology term from an id, always checking against the remote
         """
+        # organ_region currently uses single a non-OLS ontology, the Allen Institute's Mouse Brain Atlas (MBA)
+        # MBA was originally downloaded April 2020 from Allen Brain Atlas data portal URL
+        # http://api.brain-map.org/api/v2/data/query.csv?criteria=model::Structure,rma::criteria,[ontology_id$eq1],rma::options[order$eq%27structures.graph_order%27][num_rows$eqall]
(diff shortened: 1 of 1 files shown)


ingest/validation/validate_metadata.py
```diff
@@ -265,6 +272,12 @@ def retrieve_ols_term(self, ontology_urls, term, property_name):
             raise RuntimeError(error_msg)
 
     def retrieve_mouse_brain_term(self, term, property_name):
+        """Determine whether ID is in mouse brain atlas (MBA) file
+        """
+        # MBA id is also the leaf entity of structure_id_path in the MBA file
```
eweitz, line 277:
```suggestion
        # MBA ID is also the leaf entity of structure_id_path in the MBA file
```

eweitz (approved):
Helpful!

eweitz:
For auditability, please add a Jira ticket ID to the PR title.

bistline (approved):

jlchang:
NCBI is having issues which are likely the root cause of the "multiprocessing.pool.MaybeEncodingError" errors for test_genomes and test_make_toy. We believe the CI errors are false positives and can safely be ignored.

PR #178, opened 2020-12-03
* SCP-2916 Add debugging statements

knapii-developments:
This PR adds debugging statements around functions surrounding Mixpanel logging. Adding debug statements will help improve visibility regarding why ingest is not logging to Mixpanel in production.

### DIFF
--- a/ingest/config.py
+++ b/ingest/config.py
@@ -1,9 +1,14 @@
 # File is responsible for defining globals and initializing them
 try:
     from mongo_connection import MongoConnection
+    from monitor import setup_logger
 except ImportError:
     from .mongo_connection import MongoConnection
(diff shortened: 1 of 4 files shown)


bistline (approved):

eweitz (approved):
Per our chat, for debugging, it'd likely help to temporarily hard-code the production Bard host URL in your development Ingest Pipeline.

Also, non-blocking: for easier searchability and less readability noise, it'd be worth cleaning up the colons in the log messages, e.g. "foo: bar" instead of "foo : bar", "foo :bar", etc.

knapii-developments:
After further review with @eweitz, ingest is actually logging to Mixpanel. The work in this PR is no longer needed.

PR #185, opened 2020-12-14
* SCP-2949 Ea fix raw count

knapii-developments:
Raw count matrices can have the same cell names as processed matrices. This Pr fixes how duplicate cell values are evaluated.

### DIFF
--- a/ingest/expression_files/expression_files.py
+++ b/ingest/expression_files/expression_files.py
@@ -103,7 +103,7 @@ class GeneExpression:
         # If study files does not have document expression_file_info
         # field, "is_raw_count_files", will not exist.:
         if "expression_file_info" in study_file_doc.keys():
-            return study_file_doc["expression_file_info"]["is_raw_counts"]
+            return study_file_doc["expression_file_info"]["is_raw_count_files"]
         else:
             return False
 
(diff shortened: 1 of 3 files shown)


ingest/expression_files/expression_files.py
```diff
@@ -103,7 +103,7 @@ def is_raw_count_file(study_id, study_file_id, client):
         # If study files does not have document expression_file_info
         # field, "is_raw_count_files", will not exist.:
         if "expression_file_info" in study_file_doc.keys():
-            return study_file_doc["expression_file_info"]["is_raw_counts"]
+            return study_file_doc["expression_file_info"]["is_raw_count_files"]
```
bistline, line 106: This is incorrect - the existing attribute of `is_raw_counts` is what's in the model.

ingest/expression_files/expression_files.py
```diff
@@ -207,7 +213,7 @@ def get_study_expression_file_ids(
         )
         if is_raw_count_files:
             QUERY["$and"].append(
-                {"expression_file_info.is_raw_counts": is_raw_count_files}
+                {"expression_file_info.is_raw_count_files": is_raw_count_files}
```
bistline, line 216: Same problem - the attribute is `is_raw_counts`.

tests/test_expression_files.py
```diff
@@ -210,18 +210,15 @@ def test_get_cell_names_from_study_file_id(self, mock_query_cells):
...
+            cells = GeneExpression.get_cell_names_from_study_file_id(
+                ObjectId(), ObjectId(), MagicMock()
             )
+            self.assertEqual(cells, None)
 
     def test_is_raw_count_file(self):
         client = MagicMock()
         client["study_files"].find.return_value = [
-            {"expression_file_info": {"is_raw_counts": False}}
+            {"expression_file_info": {"is_raw_count_files": False}}
```
bistline, line 221: Same - all the references in this test are incorrect.

bistline (changes requested):
The reference to `is_raw_counts` was correct and didn't need to be changed.

PR #193, opened 2020-12-15
* SCP-2949 Fix query for processed matrices

knapii-developments:
Ingest pipeline was not querying correctly to retrieve cell values for processed files. This was causing all cell values to be returned when a raw count file was ingested first. This PR fixes the issue.

### DIFF
--- a/ingest/expression_files/expression_files.py
+++ b/ingest/expression_files/expression_files.py
@@ -212,6 +212,7 @@ class GeneExpression:
                 {"expression_file_info": None},
             ]
         else:
+            # Raw counts need appendage
             QUERY["$and"].append(
                 {"expression_file_info.is_raw_counts": is_raw_count_files}
             )
--- a/ingest/expression_files/expression_files.py
+++ b/ingest/expression_files/expression_files.py
@@ -206,13 +206,12 @@ class GeneExpression:
             study_id, current_study_file_id, client
         )
         if not is_raw_count_files:
-            # Legacy data entries will not have the embedede document 'expression_file_info'
+            # Legacy data entries will not have the embeded document 'expression_file_info'
             QUERY["$or"] = [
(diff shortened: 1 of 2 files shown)


ingest/expression_files/expression_files.py
```diff
@@ -205,7 +205,13 @@ def get_study_expression_file_ids(
         is_raw_count_files = GeneExpression.is_raw_count_file(
             study_id, current_study_file_id, client
         )
-        if is_raw_count_files:
+        if not is_raw_count_files:
+            # Legacy data entries will not have the embeded document 'expression_file_info'
+            QUERY["$or"] = [
```
bistline, line 210: Are we 100% certain that this `$or` is still scoped by the rest of the query?  Meaning the `$and` and `$nor` are still honored even if the `$or` is present?

ingest/expression_files/expression_files.py
```diff
@@ -205,7 +205,13 @@ def get_study_expression_file_ids(
         is_raw_count_files = GeneExpression.is_raw_count_file(
             study_id, current_study_file_id, client
         )
-        if is_raw_count_files:
+        if not is_raw_count_files:
```
devonbush, line 208: let's flip the order of this if/else so it can be `if is_raw_counts_files` and not need the negation

bistline (approved):
Looks good, just a question about the scope of the `$or` query.

devonbush (approved):
looks fine, one style change recommended

eweitz (approved):
Glad to hear the staging hotfix technique let us short-circuit manual integration tests.

ingest/expression_files/expression_files.py
knapii-developments: 136df7f

jlchang (approved):

PR #195, opened 2020-12-16
* SCP-2844 Exclude commented lines from unsorted MTX files

knapii-developments:
The PR addresses the ```not enough values to unpack (expected 3, got 1)``` error caused by headers being present when models were being created in transform(). The error was isolated to disordered MTX files. Transform() is expecting the pointer to be where the data starts. Instead, it received the header and not the gene-barcode-expression array. This was due to sort_mtx() not passing a file whose pointer was at the top to get_gene_expression_data(). To ensure the file pointer is always at the top of the file, get_gene_expression_data() [forces the pointer to the top of the file](https://github.com/broadinstitute/scp-ingest-pipeline/blob/7a1883a9b065fce6292a135c568047d4d0e1b4eb/ingest/expression_files/mtx.py#L195) with Python's [seek method](https://www.geeksforgeeks.org/python-seek-function/).

### DIFF
--- a/ingest/expression_files/mtx.py
+++ b/ingest/expression_files/mtx.py
@@ -309,7 +309,7 @@ class MTXIngestor(GeneExpression, IngestFiles):
             # Cell names are the only data stored for raw counts.
             # Therefore, no need to determine if file is sorted or sort file.
             self.is_raw_count = True
-            self.transform()
+        self.transform()
 
     def extract_feature_barcode_matrices(self):
         """
--- a/ingest/expression_files/mtx.py
+++ b/ingest/expression_files/mtx.py
@@ -183,6 +183,10 @@ class MTXIngestor(GeneExpression, IngestFiles):
     def get_data_start_line_number(file_handler: IO) -> int:
         """ Determines what line number data starts.
 
+        Parameters:
+        ___________
+            file_handler (IO): File handler of MTX file that contains headers.
+
          Returns
          ----------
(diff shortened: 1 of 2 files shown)


bistline (approved):

ingest/expression_files/mtx.py
```diff
@@ -246,7 +248,8 @@ def sort_mtx(file_path, mtx_file_handler: IO) -> str:
 
          Returns
          ----------
-            new_file_path (str) : Full path of newly sorted MTX file
+            new_file_path (str) : Full path of newly sorted MTX file. This file does not contain original headers
```
eweitz, line 251:
```suggestion
            new_file_path (str): Full path of newly sorted MTX file. This file does not contain original headers
```

jlchang (approved):
Looks good!

eweitz (approved):

devonbush (approved):
was the change to the test data file intentional?  I don't understand the reason why those lines were removed

knapii-developments:
@devonbush For some reason it had too many values. I did a comparison between the 2 sorted and unsorted matrices and found out the expected had 1 too many valus for the 80th gene.
``` egrep "^80" data/mtx/AB_toy_data_toy.unsorted_mtx.mtx | wc -l
53
egrep "^80" data/mtx/AB_toy_data_toy.matrix.mtx | wc -l
      53
```

PR #197, opened 2020-12-21
* SCP-2975 Add validation to check cell names in cluster file exist in metadata file

knapii-developments:
The PR adds validation to ensure that cell names in the cluster file exist in the metadata file. Without this validation, subsampling will hang. As a result, the user will not get the email regarding the parsing status, nor can they delete the metadata file via the UI. Instead, they'll have to reach out for support, and a member of the team will have to delete the file manually. Now, if there is no intersection, an error will be thrown, and the user will be notified that the intersection does not exist. This will cause Ingest Pipeline to fail, and normal rails clean up will proceed.


ingest/subsample.py
```diff
@@ -28,6 +28,16 @@ def __init__(self, cluster_file, cell_metadata_file=None):
...
 
+    @staticmethod
+    def get_annot_names(df_headers: List):
+        """Returns annotation names without annotation types"""
+        return [header[0] for header in df_headers]
+
+    @staticmethod
+    def has_cells_in_metadata_file(metadata_cells, cluster_cells):
+        """Checks if cells in cluster are in metadata cells"""
+        return set(cluster_cells).issubset(set(metadata_cells))
```
eweitz, line 39: Nice.

tests/test_subsample.py
```diff
@@ -70,17 +70,17 @@ def test_subsample(self):
                 self.assertEqual(
                     len(subsampled_values),
                     sample_size,
-                    f'Amount of values is incorrect. Expecting a sample size of {sample_size}',
+                    f"Amount of values is incorrect. Expecting a sample size of {sample_size}",
```
eweitz, line 73: Tangentially: "number of" is preferable to "amount of" here, because "values" is a count noun (how many) instead of a mass noun (how much).  More [here](http://crosstalk.cell.com/blog/the-difference-between-number-and-amount.).

```suggestion
                    f"Number of values is incorrect. Expecting a sample size of {sample_size}",
```

eweitz (approved):
Looks good!  I suggested a non-blocking error message refinement.

ingest/ingest_pipeline.py
```diff
@@ -400,16 +399,26 @@ def subsample(self):
...
             try:
-                subsample.prepare_cell_metadata()
-                for data in subsample.subsample("study"):
-                    load_status = self.load_subsample(
-                        Clusters.COLLECTION_NAME,
-                        data,
-                        subsample.set_data_array,
-                        "study",
+                # Annotation names for cluster file
+                annot_names = SubSample.get_annot_names(subsample.annot_column_headers)
```
devonbush, line 403: I don't understand what annotation names have to do with checking unique cells.  the 'annotation names' are the cluster file column headers, (e.g. 'Intensity', 'subgroup').  Do you mean cluser_file_cell_names here?

devonbush (changes requested):
Nice tests.  Three issues:
1. If someone deletes their metadata file, and then re-uploads a new metadata file, do we need to check that it contains all the cells in their cluster files?
2. Do we need to update the Study file upload/edit UI to clarify that users MUST upload a metadata file before uploading a cluster file?  This has workflow implications, since now users have to produce a convention-compliant metadata file before they can upload any clustering.  Let's run this by Vicky
3. I think one of your variables is misnamed--I *think* it's doing the right thing, just mislabeled.

ingest/ingest_pipeline.py
knapii-developments: This was actually a problem with context switching. I was just working on expression files where the cell names are the first lines. Made the update in d2979c8

devonbush (approved):

PR #201, opened 2021-01-15
Relax cell name validation for concurrent expression matrix parses (SCP-3009, SCP-3030)

bistline:
If a user uploads multiple expression matrices of the same "type" (e.g. raw/processed), there is currently a race condition with validating cell names that can cause both matrix files to fail ingest (although sometimes the second matrix will pass validation).  This is because `GeneExpression.get_cell_names_from_study_file_id` will see the presence of the other matrix files, but will not be able to find any cell name `DataArray` documents for these fails.  This throws a `ValueError` and fails the entire job.  The current workaround is to wait to upload the second matrix until the first has succeeded, but this is completely opaque to the end user.

This update removes that clause, and adds an upstream check to ensure that cells are parsed from any given matrix (a `ValueError` is raised if matrices don't have cell names: SCP-3030). When combined with work on SCP-2971 (which will stagger ingesting matrix files of the same type by waiting until the previous matrix has validated) this will allow users to upload 2+ matrix files of the same type concurrently.  End users should not notice any significant increase in parse times as validating and storing cell names is a relatively quick process, even for large matrix files.

This PR satisfies SCP-3009 and SCP-3030.

### DIFF
--- a/ingest/expression_files/expression_files.py
+++ b/ingest/expression_files/expression_files.py
@@ -148,23 +148,17 @@ class GeneExpression:
         query_results: List[Dict] = GeneExpression.query_cells(
             study_id, client, additional_query_kwargs
         )
-        # The query did not return results
+
+        # there are no results so return an empty array
         if not query_results:
-            # check if the other expression matrices are "parsing", in which case ignore
-            for matrix_file_id in study_files_ids:
-                matrix_file = list(client['study_files'].find({"_id": matrix_file_id})).pop()
-                if matrix_file['parse_status'] == 'parsing':
-                    continue
-                else:
-                    raise ValueError(
-                        "There are study files that do not have cell names associated with them"
-                    )
-        # Flatten query results
-        existing_cells = [
-            values
-            for cell_values in query_results
-            for values in cell_values.get("values")
-        ]
+            existing_cells = []
+        else:
+            # Flatten query results
+            existing_cells = [
+                values
+                for cell_values in query_results
+                for values in cell_values.get("values")
+            ]
         return existing_cells
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


eweitz (approved):

ingest/expression_files/expression_files.py
```diff
@@ -150,9 +150,15 @@ def get_cell_names_from_study_file_id(study_id, study_file_id, client):
...
-            raise ValueError(
-                "There are study files that do not have cell names associated with them"
-            )
+            # check if the other expression matrices are "parsing", in which case ignore
+            for matrix_file_id in study_files_ids:
+                matrix_file = list(client['study_files'].find({"_id": matrix_file_id})).pop()
+                if matrix_file['parse_status'] == 'parsing':
+                    continue
+                else:
+                    raise ValueError(
```
devonbush, line 159: now that we're adding the 'one parse at a time' restriction, do we still want to raise this error?  It seems like ingest pipeline shouldn't be in the business of throwing errors based on previously ingested files.  If we are forbidding expression matrices with zero cells, that check should happen at ingest time

devonbush (changes requested):
looks like it works, just one question about whether we can remove a section for simplicity

ingest/expression_files/expression_files.py
bistline: I don't have any objection to that except that if we remove it, we should add a validation somewhere else that would throw an error if we don't have any cell names to store (which is what this is checking in reverse).  The only reason that we're instituting the "one parse at a time" rule for expression matrices of the same type is to ensure that we can validate cells across all matrices, and concurrent parses cannot do this.

bistline: 2bb3ba5

eweitz (approved):

devonbush (approved):

PR #203, opened 2021-01-25
* SCP-2763 Include user identifier in logs from Ingest Pipeline to Bard

knapii-developments:
This PR adds user identifier in logs from Ingest Pipeline to Bard. The user ID is a unique UUID that comes from the [metrics_user_id field](https://github.com/broadinstitute/single_cell_portal_core/blob/c727832d699bbf29b897e9be52887be9c3fc3cd1/app/models/user.rb#L99) in the User model. Rails send this ID via the optional parameter ingest parameter `----user-metrics-uuid` in ingest pipeline. There's a [check](https://github.com/broadinstitute/scp-ingest-pipeline/blob/778a5bf44d1c4b5f9cf24e8e21541af54a2df876/ingest/cli_parser.py#L254) that validates the UUID. If a UUID is not supplied, the default ID is used. The default ID should technically never be used since Ingest Pipeline can only be instantiated by a signed in user. 

Here is a screenshot of user 1:
Screen Shot 2021-01-25 at 4 30 06 PM

User 2:
Screen Shot 2021-01-26 at 10 50 17 AM

This PR needs to be released with [pull #877](https://github.com/broadinstitute/single_cell_portal_core/pull/877)

### DIFF
--- a/ingest/cli_parser.py
+++ b/ingest/cli_parser.py
@@ -255,8 +255,10 @@ def is_valid_uuid(value):
     import uuid
 
     try:
-        if not value:
+        if value:
             uuid.UUID(value)
-        return value
+            return value
+        else:
+            return None
     except ValueError as e:
         raise argparse.ArgumentTypeError(e)
--- a/ingest/cli_parser.py
+++ b/ingest/cli_parser.py
@@ -20,7 +20,7 @@ def bq_dataset_exists(dataset):
         bigquery_client.get_dataset(dataset_ref)
         exists = True
     except NotFound:
-        print(f'Dataset {dataset} not found')
+        print(f"Dataset {dataset} not found")
     return exists
 
 
@@ -33,7 +33,7 @@ def bq_table_exists(dataset, table):
         bigquery_client.get_table(table_ref)
         exists = True
     except NotFound:
-        print(f'Dataset {table} not found')
+        print(f"Dataset {table} not found")
     return exists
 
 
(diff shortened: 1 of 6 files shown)


eweitz (approved):
Looks good!  Thanks for including the helpful screenshots.

I suggest a few maintainability improvements; no blockers.

ingest/config.py
```diff
@@ -34,8 +35,12 @@ def get_metric_properties():
 class MetricProperties:
     """Sets default properties for Mixpanel events"""
 
-    def __init__(self, study, study_file):
+    USER_ID = "2f30ec50-a04d-4d43-8fd1-b136a2045079"
+
+    def __init__(self, study, study_file, user_uuid=None):
+        distinct_id = user_uuid if user_uuid else MetricProperties.USER_ID
```
eweitz, line 41: Rails refers to this `user_uuid` as `metrics_uuid`.  Let's do similarly:

```suggestion
    def __init__(self, study, study_file, metrics_uuid=None):
        distinct_id = metrics_uuid if metrics_uuid else MetricProperties.USER_ID
```

(The slightly expanded form `user_metrics_uuid` seems OK, but if shortening it's worth being consistent with Rails.)

tests/test_metrics_service.py
```diff
@@ -53,6 +61,7 @@ class MetricsServiceTestCase(unittest.TestCase):
     )
     def test_log(self, mock_execute_ingest, mock_MONGO_CONNECTION, mock_post_event):
         mock_MONGO_CONNECTION._client = MetricsServiceTestCase.client_mock
+        # Test when user uuid is not provided
```
eweitz, line 64: More consistent with Rails:

```suggestion
        # Test when metrics uuid is not provided
```

tests/test_metrics_service.py
```diff
@@ -81,3 +90,18 @@ def test_log(self, mock_execute_ingest, mock_MONGO_CONNECTION, mock_post_event):
         metrics_model = config.get_metric_properties()
         # Log Mixpanel events
         MetricsService.log(config.get_parent_event_name(), metrics_model)
+
+        # Test when user uuid is provided
```
eweitz, line 94: More consistent with Rails:

```suggestion
        # Test when metrics uuid is provided
```

bistline (approved):

ingest/cli_parser.py
```diff
@@ -106,6 +106,13 @@ def create_parser():
         "--profile-memory", action="store_true", help=profile_memory_text
     )
 
+    parser.add_argument(
+        "--user-metrics-uuid",
+        required=False,
+        type=is_valid_uuid,
```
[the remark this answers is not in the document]
devonbush: I don't feel strongly about this, but, if it were me, I wouldn't validate that this is a UUID.  We're essentially just saying "give me a string that I'll pass along for you to a third-party".  Since ingest-pipeline itself doesn't care about the string itself, requiring it to be a uuid can just make the code more brittle to change, and unduly sensitive to how various languages format guids

ingest/config.py
```diff
@@ -1,5 +1,6 @@
 # File is responsible for defining globals and initializing them
 try:
+
```
devonbush, line 3: this seems like extra whitespace

devonbush (approved):
style nits, but overall good

PR #205, opened 2021-02-04
* SCP-3069 Store numeric arrays as group annotation types

knapii-developments:
Stored array-valued annotations that Mongo treated as “numeric” were plotted as NaN. To allow these annotations to be visualized in SCP, we should treat the array-based numeric data as a single string, and process the annotations as though they were regular group-based metadata entries (including storing all unique values).

From data file [valid_array_v2_1.2.txt](https://github.com/broadinstitute/scp-ingest-pipeline/blob/development/tests/data/valid_array_v2.1.2.txt), the numeric array model, 'disease__time_since_onset', will look as such:
```
{'name': 'disease__time_since_onset', 'annotation_type': 'group', 'values': [], 
'study_file_id': ObjectId('5dd5ae25421aa910a723a337'), 
'study_id': ObjectId('5d276a50421aa9117c982845')}
```

**Notice how the stored 'values', as in production, are treated as normal numeric annotations.**

Tests for the PR are captured in SCP-3082.

### DIFF
--- a/ingest/cell_metadata.py
+++ b/ingest/cell_metadata.py
@@ -49,6 +49,7 @@ class CellMetadata(Annotations):
         self.ontology = defaultdict(lambda: defaultdict(list))
         self.ontology_label = dict()
         self.cells = []
+        self.numeric_array_columns = {}
 
     # This model pertains to columns from cell metadata files
     @dataclass
@@ -67,6 +68,10 @@ class CellMetadata(Annotations):
         """ Method for ingesting cell metadata files."""
         for metadata_model in self.transform():
             yield metadata_model
+
+    def update_array_numeric_columns(self, annotation_name):
+        if not self.numeric_array_columns.values():
+            self.numeric_array_columns[annotation_name] = True
 
     # Will evolve to do cross file validation
     def validate(self, validate_against_convention=False):
@@ -105,12 +110,19 @@ class CellMetadata(Annotations):
         for annot_header in self.file.columns:
             annot_name = annot_header[0]
             annot_type = annot_header[1]
+            # When file is conventional and contain's numeric arrays
+            # the annotation type is changed to group for visualization purposes
+            stored_mongo_annot_type: str = (
+                annot_type
+                if not self.numeric_array_columns.get(annot_name)
+                else "group"
+            )
             yield AnnotationModel(
                 annot_header,
                 self.Model(
                     {
                         "name": annot_name,
-                        "annotation_type": annot_type,
+                        "annotation_type": stored_mongo_annot_type,
                         # unique values from "group" type annotations else []
                         "values": list(self.file[annot_header].unique())
                         if annot_type == "group"
@@ -122,7 +134,6 @@ class CellMetadata(Annotations):
             )
 
     def set_data_array(self, linear_data_id: str, annot_header: str):
-        """Builds data array"""
         data_array_attrs = locals()
         del data_array_attrs["annot_header"]
         del data_array_attrs["self"]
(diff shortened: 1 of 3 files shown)


bistline (changes requested):
I have a concern that while these files will ingest, they will not visualize based off of this comment: **Notice how the stored 'values', as in production, are treated as normal numeric arrays.**

The entries found in `values` for group-based annotations are in two main places: determining whether or not a particular annotation can be visualized (between 2 & 200 unique values), and for initializing all of the Plotly trace entries when rendering any visualization.  If we are ingesting these files and marking them as "group" without doing anything else, then they will not appear in the "Annotation" dropdown menu for selecting different annotations to visualize.

We will either need to store the unique values - in which case everything will work as normal - or some other mechanism must be devised for visualizing these particular annotations.

jlchang:
> I have a concern that while these files will ingest, they will not visualize based off of this comment: **Notice how the stored 'values', as in production, are treated as normal numeric arrays.**
> 
> The entries found in `values` for group-based annotations are in two main places: determining whether or not a particular annotation can be visualized (between 2 & 200 unique values), and for initializing all of the Plotly trace entries when rendering any visualization. If we are ingesting these files and marking them as "group" without doing anything else, then they will not appear in the "Annotation" dropdown menu for selecting different annotations to visualize.
> 
> We will either need to store the unique values - in which case everything will work as normal - or some other mechanism must be devised for visualizing these particular annotations.

I'm surprised we'd store the values as a numeric array. I'd assume we would store them as an array of strings (specifically, strings that have numbers and zero or more small amount of pipes in the string). For [valid_array_v2.1.2.txt](https://github.com/broadinstitute/scp-ingest-pipeline/blob/development/tests/data/valid_array_v2.1.2.txt), I would expect the array of values in Mongo to be exactly what Eno saw when she looked in Mongo on Friday
Screen Shot 2021-01-29 at 5 04 37 PM
(ie "values" : [ "12|2", "1", "24|2", "36|3|1", "0" ])

bistline:
> (ie "values" : [ "12|2", "1", "24|2", "36|3|1", "0" ])

Sorry - I'm proposing exactly that: storing the unique values as an array of strings, like you've shown.

knapii-developments:
> > (ie "values" : [ "12|2", "1", "24|2", "36|3|1", "0" ])
> 
> Sorry - I'm proposing exactly that: storing the unique values as an array of strings, like you've shown.

Just checking that we're all on the same page,  the model shown in the description is the cell metadata model and not the data array model.

jlchang:
> > > (ie "values" : [ "12|2", "1", "24|2", "36|3|1", "0" ])
> > 
> > 
> > Sorry - I'm proposing exactly that: storing the unique values as an array of strings, like you've shown.
> 
> Just checking that we're all on the same page, the model shown in the description is the cell metadata model and not the data array model.

Is there some way to re-word the opening PR description so future me doesn't mis-interpret again?

bistline:
How about this:

> Stored array-valued annotations that Mongo treated as “numeric” were plotted as NaN. To allow these annotations to be visualized in SCP, we should treat the array-based numeric data as a single string, and process the annotations as though they were regular group-based metadata entries (including storing all unique values).

bistline (approved):

ingest/cell_metadata.py
```diff
@@ -122,8 +135,8 @@ def transform(self):
             )
 
     def set_data_array(self, linear_data_id: str, annot_header: str):
-        """Builds data array"""
-        data_array_attrs = locals()
+
```
eweitz, line 138: A comment here would help maintainability.

eweitz (approved):
Looks good.  I suggested a typo fix and a maintainability refinement -- no blockers.

devonbush (approved):

PR #206, opened 2021-02-24
* SCP-2909 Resolve "batch ops error" with retry using new random document ID

knapii-developments:
This PR attempts to resolve the batch-ops error due to duplicate IDs by assigning another random ID and trying to re-insert. 

To test:
Find a file that already present in your Mongo instance.
On local instance of scp-ingest-pipeline try and ingest the same file with the same study and study file ID.
In ingest_files add ```self._id = < _id value in document you're truing to duplicate>```
In logs, you'll see that the batch ops error occurred. However, you'll see in Mongo the inserted document has a different id.

### DIFF
--- a/ingest/mongo_connection.py
+++ b/ingest/mongo_connection.py
@@ -1,6 +1,7 @@
 import os
 import functools
 import time
+from bson.objectid import ObjectId
 from pymongo import MongoClient
 from pymongo.errors import AutoReconnect, BulkWriteError
 
@@ -65,6 +66,7 @@ def graceful_auto_reconnect(mongo_op_func):
 
     @functools.wraps(mongo_op_func)
     def wrapper(*args, **kwargs):
+        args = list(args)
         for attempt in range(MAX_ATTEMPTS):
             try:
                 return mongo_op_func(*args, **kwargs)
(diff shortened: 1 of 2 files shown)


ingest/mongo_connection.py
```diff
@@ -75,13 +77,25 @@ def wrapper(*args, **kwargs):
...
+                        args[0] = reassign_id(args[0])
                     retry(attempt)
                 else:
                     dev_logger.debug(str(bwe.details))
                     raise bwe
 
     return wrapper
+
+
+def reassign_id(documents):
```
bistline, line 97: Technically, the error code `11000` is `duplicate key error index`, which means that an index is invalid because we're violating a constraint, usually for uniqueness.  I think all the instances we've run across right now are for the main document ID, but it is possible that there are other unique indexes that are throwing the error.  For instance, there are 3 other indexes for `DataArray` documents, one of which is a unique 10-value compound index that could also throw the same error, but it would not be for the main document ID.

Not a blocking concern for now, but something we should be aware of.

bistline (approved):

devonbush (approved):

PR #209, opened 2021-03-05
* SCP-3082 Tests for cell metadata

knapii-developments:
Cell metadata has gotten updated to fix bugs. This PR add test cases as defined in is [documentation](https://docs.google.com/document/d/1HipKN9YwjDemDyh3XIK9wZ37crtqEQB93QCrejkedSY/edit).


ingest/validation/validate_metadata.py
```diff
@@ -211,9 +213,15 @@ def retrieve_ols_term(self, ontology_urls, term, property_name):
...
             ontology_shortname, term_id = re.split("[_:]", term)
         except (ValueError, TypeError):
-            raise ValueError(
-                f'{property_name}: Could not parse provided ontology id, "{term}"'
-            )
+            msg = f'{property_name}: Could not parse provided ontology id, "{term}".'
+            if type == "array":
+                if "|" not in term:
+                    msg += (
+                        f" There is only one array value, for ontology id, '{term}.' "
```
bistline, line 220: I like that we're specifically calling this out now - a good enhancement.

ingest/validation/validate_metadata.py
```diff
@@ -559,6 +568,20 @@ def collect_cell_for_ontology(
...
+        return (
+            # check for missing values
+            ontology_label not in updated_row
+            # check for empty cells
+            or value_is_nan(cell_name)
+            # check for empty cells
+            or is_empty_string(cell_name)
+        )
+
+    # check for missing values
```
bistline, line 581: Seems like copy/paste detritus here.

ingest/validation/validate_metadata.py
```diff
@@ -713,12 +735,15 @@ def value_is_nan(value):
     """
     # We coerces nan to string for group annotations (SCP-2545)
     # string value 'nan' should also be considered value_is_nan=True
-    if value == "nan":
+    if str(value).lower() == "n/a":
```
bistline, line 738: Do we also need to account for `"na"` as well?

ingest/validation/validate_metadata.py
```diff
@@ -999,14 +1051,18 @@ def validate_collected_ontology_data(metadata, convention):
...
         ontology_urls = convention["properties"][property_name]["ontology"].split(",")
 
         for ontology_info in metadata.ontology[property_name].keys():
             ontology_id, ontology_label = ontology_info
+
             try:
+                attribute_type = convention["properties"][property_name]["type"]
                 matched_label_for_id = retriever.retrieve_ontology_term_label(
-                    ontology_id, property_name, convention
+                    ontology_id, property_name, convention, attribute_type
```
bistline, line 1063: Minor nit - some places we call this variable `type`, and other places `attribute_type`.  Standardizing would help readability.  I personally like `attribute_type` as it is more informative.

tests/test_cell_metadata.py
```diff
@@ -1,8 +1,16 @@
 import sys
 import unittest
+import json
+
+from mock_data.annotation.metadata.convention.valid_array_v2_1_2 import (
```
bistline, line 5: How come this is done via an import?  Is it because `valid_array_v2_1_2.py` has `ObjectID()` values?  Couldn't we just have static values for those?

tests/test_validate_metadata.py
```diff
@@ -530,35 +578,71 @@ def test_invalid_ontology_content(self):
...
 
-    def test_valid_array_content(self):
-        """array-based metadata should conform to convention requirements
+    def test_content(self):
+        """Array-based metadata should conform to convention requirements
         """
-        args = "--convention ../schema/alexandria_convention/alexandria_convention_schema.json ../tests/data/valid_array_v2.1.2.txt"
-        metadata, convention = self.setup_metadata(args)
+
+        def set_up_test(test_file_name, ref_file_name):
```
bistline, line 585: I like how you abstracted this.

bistline (approved):
Just a few questions/style nits.  I'm a little surprised the coverage only went up by .04% - I would have expected more.  Nonetheless, nice work on this rather large effort.

ingest/validation/validate_metadata.py
```diff
@@ -154,7 +154,7 @@ class OntologyRetriever:
     cached_ontologies = {}
     cached_terms = {}
 
-    def retrieve_ontology_term_label(self, term, property_name, convention):
+    def retrieve_ontology_term_label(self, term, property_name, convention, type):
```
devonbush, line 157: this could use a comment about what these arguments are

ingest/validation/validate_metadata.py
```diff
@@ -171,12 +171,14 @@ def retrieve_ontology_term_label(self, term, property_name, convention):
...
             ] = self.retrieve_ontology_term_label_remote(
-                term, property_name, ontology_urls
+                term, property_name, ontology_urls, type
             )
 
         return self.cached_terms[property_name][term]
 
-    def retrieve_ontology_term_label_remote(self, term, property_name, ontology_urls):
+    def retrieve_ontology_term_label_remote(
+        self, term, property_name, ontology_urls, type
```
devonbush, line 180: same -- what is 'type'?

ingest/validation/validate_metadata.py
```diff
@@ -211,9 +213,15 @@ def retrieve_ols_term(self, ontology_urls, term, property_name):
...
         except (ValueError, TypeError):
-            raise ValueError(
-                f'{property_name}: Could not parse provided ontology id, "{term}"'
-            )
+            msg = f'{property_name}: Could not parse provided ontology id, "{term}".'
+            if type == "array":
+                if "|" not in term:
+                    msg += (
+                        f" There is only one array value, for ontology id, '{term}.' "
+                        "If unexpected, multiple values must be pipe ('|') delimited."
```
devonbush, line 221: what does 'if unexpected' mean here?  

would 'To specify multiple values, make them pipe ('|') delimited' ? be clearer?

ingest/validation/validate_metadata.py
devonbush: I don't understand how this check is operating.  this method is supposed to only handle single terms, right?  So if a term with a pipe got passed to this method, doesn't that mean our upstream parsing failed to recognize it as an array?

devonbush (changes requested):
This seems like a lot of unit-level changes to logic, but with only integration tests at the whole-file level, it makes it extremely hard to confirm the behavior.  I think these need unit tests so that it's very clear what edge cases we can handle (e.g. what about numeric array columns with a mix of single and multiple values)?  
Happy to pair with you on how to create those unit tests -- I'm looking for tests similar to test_validate_metadata.py

ingest/validation/validate_metadata.py
knapii-developments: We actually allow single value arrays. This is shown in the valid array test file [here](https://github.com/broadinstitute/scp-ingest-pipeline/blob/2bcddddf4ecc6a508be47a198d9751316cbe30c9/tests/data/annotation/metadata/convention/valid_array_v2.1.2.txt#L7). This check is here due to this issue stated in [this comment](https://github.com/broadinstitute/scp-ingest-pipeline/pull/209/files#r591691638)

ingest/validation/validate_metadata.py
knapii-developments: This is taken care of using the isnan function in line [742](https://github.com/broadinstitute/scp-ingest-pipeline/blob/2bcddddf4ecc6a508be47a198d9751316cbe30c9/ingest/validation/validate_metadata.py#L742).

devonbush (approved):

PR #212, opened 2021-06-04
Release 1.11.0

bistline:


eweitz (changes requested):
This updates metadata validation, which is [used in our public CLI](https://github.com/broadinstitute/single_cell_portal/blob/master/scripts/manage_study.py#L41) (i.e., `manage-study`), so let's [update the version in our setup.py to 1.11.0](https://github.com/broadinstitute/scp-ingest-pipeline/blob/development/setup.py#L8).

We should also release both `scp-ingest-pipeline` and `single-cell-portal` to PyPI, as described [here](https://docs.google.com/document/d/1vJz6hXdjAu8LBAfIS4oMy7yLIUX177gU1H7BxTPPzAU/edit#heading=h.tjhhwe51jlwg) and [here](https://docs.google.com/document/d/1IJYGz0IiaFKH5O0NoExrPKUessYhgYQ5LokkTdZnDks/edit).

eweitz (approved):

bistline:
Given the problems that have arisen in the `single-cell-portal` repo regarding imports from `scp-ingest-pipeline`, we are deferring creating the `single-cell-portal` PyPI package as a part of this release.  Fixing the import issues is now tracked in SCP-3415.

jlchang (approved):
Wasn't sure if we had captured anywhere that a 1.11.0 instance was pushed to PyPI. Jon - has 1.11.0 been yanked/deleted?

So if we're able to resolve the imports issues (SCP-3415), we'll  want to push a 1.11.1 version to PyPI.

PR #215, opened 2021-09-24
Release 1.12.0

bistline:


jlchang (approved):

eweitz (approved):
Ideally this would also update [`version` in `setup.py`](https://github.com/broadinstitute/scp-ingest-pipeline/blob/development/setup.py#L8), but as that's only used for our currently-unsupported public CLI, I don't think lacking that is a blocking issue.

PR #216, opened 2021-10-12
Relocate IngestPipeline method about metadata convention to CellMetadata Class (SCP-3632)

jlchang:
Tests in single_cell_portal repo fail due to Classes in the repo that were refactored earlier this year. Methods used for tests in single_cell_portal are now obsolete. To use scp-ingest-pipeline as-is in the tests, an IngestPipeline object would need to be instantiated to use the conforms_to_metadata_convention method, which is more complex than instantiating a CellMetadata object which is what the current, broken tests do. Logically, conforms_to_metadata_convention should be a CellMetadata method because none of the other ingest file types require the metadata convention. To ease the repair of tests in the single_cell_portal repo, this PR moves the conforms_to_metadata_convention method to the Cell Metadata Class.

To test (is kinda finicky right now, sorry!)
pre-test: find a study in your dev instance and get a valid study_id and study_file_id
ensure your python environment is up to date with requirements.txt in scp-ingest-pipeline repo
(not sure if this works differently if your gcloud config is set to your service account or your Broad identity, I used my Broad identity)
Have your mondoDB credentials set up as environment variables (Let me know if this isn't documented already, I think Eno made a doc somewhere. I have a script that set's it up for me using vault commands)

Run the following 
python ingest_pipeline.py --study-id <your study_id> --study-file-id <your  study-file-id> ingest_cell_metadata --cell-metadata-file <path/to/repo>/scp-ingest-pipeline/tests/data/annotation/metadata/convention/valid_array_v2.1.2.txt --study-accession <your study accession> --ingest-cell-metadata --validate-convention

Expect [many warning messages](https://docs.google.com/document/d/1baT6Rn1TNwccele56vIstvTjHnRgacbTkmMCPF5CIBI/edit#heading=h.fk1rtq1m2hm) but no errors.

This work supports SCP-3631 and corrects the import errors that blocked SCP-3414.

### DIFF
--- a/ingest/cell_metadata.py
+++ b/ingest/cell_metadata.py
@@ -123,7 +123,8 @@ class CellMetadata(Annotations):
             self.kwargs["validate_convention"] is not None
             and self.kwargs["validate_convention"]
         ):
-            if self.kwargs["bq_dataset"] and self.kwargs["bq_table"]:
+            import_to_bq = self.kwargs["bq_dataset"] and self.kwargs["bq_table"]
+            validate_input_metadata(self, convention, bq_json=import_to_bq)
                 validate_input_metadata(self, convention, bq_json=True)
             else:
                 validate_input_metadata(self, convention)
--- a/ingest/cell_metadata.py
+++ b/ingest/cell_metadata.py
@@ -119,15 +119,9 @@ class CellMetadata(Annotations):
         convention_file_object = IngestFiles(self.JSON_CONVENTION, ["application/json"])
         json_file = convention_file_object.open_file(self.JSON_CONVENTION)
         convention = json.load(json_file)
-        if (
-            self.kwargs["validate_convention"] is not None
-            and self.kwargs["validate_convention"]
-        ):
-            import_to_bq = self.kwargs["bq_dataset"] and self.kwargs["bq_table"]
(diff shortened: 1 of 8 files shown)


setup.py
```diff
@@ -5,7 +5,7 @@
 
 setup(
     name="scp-ingest-pipeline",
-    version="1.11.0",
+    version="1.11.1",
```
eweitz, line 8: [Current version](https://github.com/broadinstitute/scp-ingest-pipeline/releases) is 1.12.0, so this should be higher than that.

ingest/cell_metadata.py
```diff
@@ -103,6 +114,23 @@ def validate_header_for_coordinate_values(self):
             self.store_validation_issue("error", "format", msg)
             return False
 
+    def conforms_to_metadata_convention(self):
+        """ Determines if cell metadata file follows metadata convention"""
+        convention_file_object = IngestFiles(self.JSON_CONVENTION, ["application/json"])
```
eweitz, line 119: Out of scope, but it'd be nice if we define default MIME types for given extensions (e.g. (`"json": ["application/json"]`) in only one place, so we could typically omit second parameters like this `["application/json"]`.

bistline (approved):
Thanks for figuring this out finally!

PR #221, opened 2021-11-10
Handle very large group-based metadata annotations in ingest (SCP-3761)

jlchang:
Metadata files may have group annotations with many unique values - they may have worth for analysis (or it's just easier for the study owner if they do not need to be removed from the analysis file). but >200 values has not proven to be useful for visualization, therefore, the values do not need to be stored in Mongo. This PR avoids loading to Mongo any group annotations in the metadata file that have >200 unique values.

To test, set your local instance "Ingest Pipeline Docker Image" configuration to use the following image:
gcr.io/broad-singlecellportal-staging/scp-ingest-jlc_skip_lg_grp_metadata:a72f76b 

and upload the metadata file found at:
https://github.com/broadinstitute/scp-ingest-pipeline/blob/jlc_skip_lg_grp_metadata/tests/data/annotation/metadata/convention/lg_grp_metadata_to_skip.txt

In the Study details page the metadata 'barcodekey' is listed and has no values.

In the ingest email, barcodekey is listed as a entry but has no associated values in the listing:
`barcodekey: group`

This addresses SCP-3761


ingest/cell_metadata.py
```diff
@@ -141,21 +150,34 @@ def transform(self):
...
-                        "study_id": self.study_id,
-                    }
-                ),
-            )
+
+            group = True if annot_type == "group" else False
+            # should not store annotations with >200 unique values for viz
+            # annot_header is the column of data, which includes name and type
+            # large is any annotation with more than 200 + 2 unique values
+            large = True if len(list(self.file[annot_header].unique())) > 202 else False
```
eweitz, line 158: The `True if... else False` construct is redundant -- the result of the inner expression is itself a Boolean, and should simply be assigned as such.

Also, unpacking this to have a `unique_values` variable would help readability farther down.

```suggestion
            group = annot_type == "group"
            
            # should not store annotations with >200 unique values for viz
            # annot_header is the column of data, which includes name and type
            # large is any annotation with more than 200 + 2 unique values
            unique_values = list(self.file[annot_header].unique())
            large = len(unique_values) > 202
```

ingest/cell_metadata.py
```diff
@@ -141,21 +150,34 @@ def transform(self):
...
+
+            if large and group:
+                skip_msg = f"\'{annot_name}\' not stored for visualization - too many unique values "
```
eweitz, line 161:
```suggestion
                skip_msg = f"\'{annot_name}\' not stored for visualization - too many unique values"
```

ingest/cell_metadata.py
```diff
@@ -141,21 +150,34 @@ def transform(self):
...
+                yield AnnotationModel(
+                    annot_header,
+                    self.Model(
+                        {
+                            "name": annot_name,
+                            "annotation_type": stored_mongo_annot_type,
+                            # unique values from "group" type annotations else []
+                            "values": list(self.file[annot_header].unique())
+                            if annot_type == "group"
+                            else [],
```
eweitz, line 175: Reuse variables from above for readability, and to avoid the dreaded multi-line ternary.

```suggestion
                            "values": unique_values if group else [],
```

tests/test_cell_metadata.py
```diff
@@ -105,3 +105,29 @@ def test_transfrom(self):
             model_name = model["name"]
             expect_model = valid_array_v2_1_2_models["cell_metadata_models"][model_name]
             self.assertEqual(model, expect_model)
+
+    def test_skip_lg_group_transform(self):
+        # metadata "barcodekey" - 250 unique values, should not transform
+        # metadata "scale" - 200 unique values , should transform
+        cm = CellMetadata(
+            "../tests/data/annotation/metadata/convention/lg_grp_metadata_to_skip.txt",
```
eweitz, line 113: Blocker: `lg` would confuse me if I came across it outside this PR.  For readability, let's not disemvowel here.

```suggestion
    def test_skip_large_group_transform(self):
        # metadata "barcodekey" - 250 unique values, should not transform
        # metadata "scale" - 200 unique values , should transform
        cm = CellMetadata(
            "../tests/data/annotation/metadata/convention/large_group_metadata_to_skip.txt",
```

(This suggestion also entails renaming the mock data file.)

eweitz (changes requested):
Good stuff!  I note a trivial blocker regarding maintainability, and a few nits.

eweitz (approved):

ingest/cell_metadata.py
jlchang: Thanks for the readability improvement with `unique_values`, very helpful.

The `True if... else False` is necessary for the `if large and group` conditional further down. Without `else False`, the variable may be unassigned and the conditional fails to evaluate successfully. Happy to discuss if I've mis-interpreted the comment.

ingest/cell_metadata.py
jlchang: Adjusting here to fit Jon's original conception - ingest all metadata, just don't store **values** for large group.

bistline (approved):
Looks good - we can merge in this release as a part of SCP-3889

PR #224, opened 2021-11-23
Reject cluster file if coordinate value of NaN detected (SCP-3575)

jlchang:
Add a check to detect if any (or all) coordinate values in a cluster file are NaN. Reject if NaN detected for X, Y or Z values.

To test, set your local instance "Ingest Pipeline Docker Image" configuration to use
`gcr.io/broad-singlecellportal-staging/scp-ingest-jlc_check_cluster_coord_not_nan:00fd4f7 `
and upload either cluster file below:
`https://raw.githubusercontent.com/broadinstitute/scp-ingest-pipeline/jlc_check_cluster_coord_not_nan/tests/data/cluster_bad_space_delimited.txt`
with this feedback in the ingest failure email:

> *** format error list:
> Missing coordinate values in x column
> Missing coordinate values in y column
> Cluster file format invalid

or
`https://raw.githubusercontent.com/broadinstitute/scp-ingest-pipeline/jlc_check_cluster_coord_not_nan/tests/data/cluster_bad_missing_coordinate_values.txt`

with this feedback in the ingest failure email:

> *** format error list:
> Missing coordinate values in y column
> Missing coordinate values in z column
> Cluster file format invalid

This PR resolves SCP-3575.

<hr>

Why did we need this check:
In performance testing megapuck data shared by the Macosko Lab (SCP1500) we found that the supplied cluster file passed ingest but would subsequently error in subsampling. The problematic file had tab-delimited header lines and space delimited values with BOTH integer and float values. This odd combination was likely treated as tab delimited but used the whole row as the cell name, resulting in X and Y having columns of NaN values (which are validly numeric in Python), allowing a false negative successful ingest.

Further testing found that cluster files with NaN for any individual coordinate value would also successfully ingest (false negative). Ingest of such data could lead to unplottable points.


ingest/clusters.py
```diff
@@ -73,7 +74,30 @@ def __init__(
...
+        for annot_name in self.headers:
+            if annot_name.lower() in ("x", "y", "z"):
+                if self.file[annot_name].isna().any().bool():
+                    valid = False
+                    msg = f"Missing coordinate values in {annot_name} column"
+                    self.store_validation_issue("error", "format", msg)
+        if valid:
+            return True
+        else:
+            return False
```
eweitz, line 100: Less redundantly:

```suggestion
        is_valid = True
        for annot_name in self.headers:
            if annot_name.lower() in ("x", "y", "z"):
                if self.file[annot_name].isna().any().bool():
                    is_valid = False
                    msg = f"Missing coordinate values in {annot_name} column"
                    self.store_validation_issue("error", "format", msg)
    return is_valid
```

tests/test_annotations.py
```diff
@@ -170,9 +170,9 @@ def test_coerce_numeric_values(self):
         dtype = cm.file.dtypes[("Average Intensity", "numeric")]
         self.assertEqual(dtype, np.float)
 
-        # Test that numeric values wer
-        # Pick a random number between 1 and amount of lines in file
-        ran_num = random.randint(1, 20)
+        # Test that numeric values were properly rounded
+        # Pick a random number to chose a line in the test file
```
eweitz, line 174:
```suggestion
        # Pick a random number to choose a line in the test file
```

eweitz (approved):
Code looks good!  I really appreciate having the example use-case context in the PR.

I note a typo, and suggest a readability refinement.

bistline (approved):
I had to try a few times, but eventually I got the file to fail with the correct error message.  The first few times it failed was with this:
```
ValueError: Unsupported file format. Allowed file types are: text/csv text/plain text/tab-separated-values
```

After inspecting the file, I noticed that it had Windows `CRLF` line endings.  I had thought that Python/pandas was tolerant of this, but I'm suspecting that perhaps it is not.  Given the prevalence of this particular error in ingest, we may want to look into line ending support more.

ingest/clusters.py
```diff
@@ -73,7 +74,27 @@ def __init__(
...
+
+    def require_X_Y_not_nan(self):
+        """
+        X Y and Z are expected to be fully populated in cluster files
+        Fail coordinate data columns that have nan values
+        """
+        is_valid = True
+        for annot_name in self.headers:
+            if annot_name.lower() in ("x", "y", "z"):
+                if self.file[annot_name].isna().any().bool():
```
ehanna4, line 93: Do we have to worry about an empty string here at all? It looks like `isna()`wouldn't count `''` as True here so just wanted to check (per these [docs](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.isna.html))

ingest/clusters.py
```diff
@@ -73,7 +74,27 @@ def __init__(
...
+                self.is_valid_format(),
+                self.validate_numeric_annots(),
+                self.require_X_Y_not_nan(),
+            ]
+        )
+
+    def require_X_Y_not_nan(self):
+        """
+        X Y and Z are expected to be fully populated in cluster files
+        Fail coordinate data columns that have nan values
```
ehanna4, line 88: tiny NIT but wondering if this should be NaN to keep the capitalization consistent?

ehanna4 (commented):
I'm not sure I have things setup locally to test this manually but the code looked good! Just had 2 small things otherwise looks nice!

ingest/clusters.py
jlchang: Great question. We use Pandas to read in the data and any empty cells (a la  missing values in a spreadsheet) are converted to NaN. If the tsv file has literal quotes with nothing between, then ingest will hit a `could not convert string to float` ValueError. I'm not planning to catch that type of error gracefully because it seems hard to generate accidentally (having quotes in a column of numeric values) and also harder to overlook visually.

Thanks for the question as it prompted me to test a case I hadn't thought of!

ingest/clusters.py
jlchang: fixed in c7c3f38

PR #234, opened 2022-02-15
Store ontology label in BigQuery for synonyms in metadata (SCP-3987)

jlchang:
Storing synonyms in BigQuery generates "clutter" in Advanced search facet filters. Previously manual cleanup has tried but synonym storage allows "clutter" to re-accumulate. This PR aims to remove the root cause of some of the "clutter" in facet filters (The other contributing cause is change from upstream ontologies, which is less frequent).

To test:

1. In your local instance, set up this docker container to perform ingest:
`gcr.io/broad-singlecellportal-staging/scp-ingest-jlc_synonym_mitigation:d7b1747`

2. ingest metadata file from single_cell_portal_core repo:
> db/seed/synthetic_studies/array_mouse_eye/metadata.tsv

3. Confirm in study details for the study, that disease__ontology_label had arrays with synonyms like:
astigmatism (disease)
cataract (disease)
glaucoma (disease)

4. Go to BigQuery via GCP console and check the stored ontology labels by running this query:
```
SELECT DISTINCT dz_ol, 
	        FROM cell_metadata_development.alexandria_convention as c,
	        UNNEST(c.disease__ontology_label) AS dz_ol
	WHERE c.study_accession = "SCPXX"
```
Identify the study in your local instance that you used for upload and replace XX with the accession

Confirm dz_ol do not have any synonyms (ie have "(disease)" appended to the term):
cataract
astigmatism
glaucoma
HIV infectious disease


ingest/validation/validate_metadata.py
```diff
@@ -1402,6 +1410,81 @@ def detect_excel_drag(metadata, convention):
...
+            # select the rows in the dataframe with entries for multivalue array (o)
+            match = [v == list(o) for v in df[metadata_name]]
+            df.loc[match, metadata_name] = df.apply(lambda x: replacement_value, axis=1)
+
+
+def replace_synonyms(metadata):
+    """
+    Update BigQuery data to store ontology labels and not synonyms
+    """
+    bq_filename = str(metadata.study_file_id) + ".json"
```
eweitz, line 1459: Study accession + file name seems easier to reason about than study file ID.

```suggestion
    bq_filename = str(metadata.study_file_id) + ".json"
```

eweitz (approved):
Code looks good!  I suggest a few maintainability refinements, nothing blocking.

ingest/validation/validate_metadata.py
```diff
@@ -1402,6 +1410,81 @@ def detect_excel_drag(metadata, convention):
     return excel_drag
 
 
+def to_1D(series):
```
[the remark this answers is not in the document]
jlchang: Thanks - I hadn't seen flatten. In python, list comprehensions are [idiomatic](https://devblogs.microsoft.com/python/idiomatic-python-comprehensions/) and usually performant.
I could use flatten for the to_1D function but list comprehensions are necessary for the functions that handle synonym replacement in array-based metadata so I'd prefer consistency across the three functions.

ingest/validation/validate_metadata.py
jlchang: I agree a study accession is more readable. However, when in "developer mode" we bypass MongoDB lookups and only have the same hard-coded dummy accession "SCPdev" regardless of what is being ingested. Developers are able to pass in variable (BSON compliant) dummy study_file_id so, for developer productivity reasons, using the study_file_id is preferable.

PR #236, opened 2022-02-28
Log ingest pipeline ValueError exceptions to Mixpanel (SCP-4139)

jlchang:
Previously ingest validation errors during "preprocess" or subsampling were not reported to Mixpanel because the mechanisms for mixpanel reporting (accessing the Annotation Class "props" dictionary) were not accessible. ValueError issues caused premature exit of the ingest_pipeline job so these failed ingests did not report as failed file-validation events. This work also covers subsampling "Cluster file has cell names that are not present in cell metadata file" errors and "Duplicate annotation header names..." ValueError that is part of Annotation file preprocessing.

In this PR, ad-hoc props for Mixpanel reporting are used to ensure that ValueError events are logged.

To test:
 
1. set your local instance to use docker image:
gcr.io/broad-singlecellportal-staging/scp-ingest-jlc_log_valueerrors:45230a2

2. upload this cluster file from ingest pipeline repo:
../tests/data/cluster_non-numeric.txt

3. confirm the following properties are in the file-validation event in Mixpanel Dev:

errorTypes: ["content:type:not-numeric"]
numErrorTypes: 1  <- when using docker image with tag 45230a2, should no longer see numError**t**ypes (lowercase t)
logger: ingest-pipeline
status: failure
trigger: upload

[optional] 4. upload a cluster file with >1000 cell entries (to trigger subsampling) and subsequently upload an unrelated metadata file (with >1000 cell entries) to trigger "Cluster file has cell names that are not present in cell metadata file" ValueError and confirm a file-validation event with property errorTypes: ["content:missing:values-across-files"].

### DIFF
--- a/ingest/ingest_pipeline.py
+++ b/ingest/ingest_pipeline.py
@@ -186,11 +186,20 @@ class IngestPipeline:
             )
         except ValueError as v:
             # Caution: recording errorTypes in this manner can clobber other collected errors.
-            # ValueErrors during file connection should stop processing so
-            # this logging approach should not lose file validation information
-            config.get_metric_properties().update(
-                {"errorTypes": ["content:type:not-numeric"]}
-            )
+            # ValueErrors during file connection indicate file cannot be processed
+            # this logging approach should not lose collected file validation information
+            if str(v).startswith("could not convert"):
+                config.get_metric_properties().update(
+                    {"errorTypes": ["content:type:not-numeric"]}
+                )
+            elif str(v).startswith("Unable to parse"):
+                config.get_metric_properties().update(
+                    {"errorTypes": ["format:cap:unique"]}
+                )
+            else:
+                config.get_metric_properties().update(
+                    {"errorTypes": ["parse:unhandled"]}
+                )
             self.report_validation("failure")
             raise ValueError(v)
 
--- a/ingest/ingest_pipeline.py
+++ b/ingest/ingest_pipeline.py
@@ -176,14 +176,23 @@ class IngestPipeline:
                 File object.
         """
         file_connections = {"cell_metadata": CellMetadata, "cluster": Clusters}
-
-        return file_connections.get(file_type)(
-            file_path,
-            self.study_id,
-            self.study_file_id,
-            tracer=self.tracer,
-            **self.kwargs,
-        )
+        try:
+            return file_connections.get(file_type)(
+                file_path,
+                self.study_id,
+                self.study_file_id,
+                tracer=self.tracer,
+                **self.kwargs,
+            )
+        except ValueError as v:
+            # Caution: recording errorTypes in this manner can clobber other collected errors.
+            # ValueErrors during file connection should stop processing so
+            # this logging approach should not lose file validation information
+            config.get_metric_properties().update(
+                {"errorTypes": ["content:type:not-numeric"]}
+            )
+            self.report_validation("failure")
+            raise ValueError(v)
 
     def insert_many(self, collection_name, documents):
         if not config.bypass_mongo_writes():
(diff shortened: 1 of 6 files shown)


ingest/ingest_pipeline.py
```diff
@@ -176,14 +176,32 @@ def initialize_file_connection(self, file_type, file_path):
...
+                config.get_metric_properties().update(
+                    {"errorTypes": ["content:type:not-numeric"], "numErrortypes": 1}
+                )
+            elif str(v).startswith("Unable to parse"):
+                config.get_metric_properties().update(
+                    {"errorTypes": ["format:cap:unique"], "numErrortypes": 1}
+                )
+            else:
+                config.get_metric_properties().update(
+                    {"errorTypes": ["parse:unhandled"], "numErrortypes": 1}
```
eweitz, line 201: Blocker: Mixpanel property names should be lower camel case to avoid downstream analytics UX bugs:

```suggestion
                    {"errorTypes": ["content:type:not-numeric"], "numErrorTypes": 1}
                )
            elif str(v).startswith("Unable to parse"):
                config.get_metric_properties().update(
                    {"errorTypes": ["format:cap:unique"], "numErrorTypes": 1}
                )
            else:
                config.get_metric_properties().update(
                    {"errorTypes": ["parse:unhandled"], "numErrorTypes": 1}
```

Two separate non-blocking issues:

First, the casing above reveals a bug in [`set_mixpanel_num`](https://github.com/broadinstitute/scp-ingest-pipeline/blob/334b09a117110237823c0c0fb55f8895c92a0c09/ingest/config.py#L63).  Using `my_string.capitalize()` capitalizes to first letter and _lowercases all subsequent letters_, which I think neither of us realized.  So, what is currently

`num_prop = "num" + prop.capitalize()`

should -- in order to achieve camel casing -- be something like

`num_prop = "num" + prop[0].upper() + prop[1:]`

This is worth ticketing for this sprint or next if you opt to not fix it as part of this PR.

Second, why omit the standard `message` item in the issue list?  If there's a trivial way to access that data in these three log points, it'd be worthwhile to enhance debuggability via Mixpanel -- e.g.  to be able to query Mixpanel on what the user reported in a support ticket, or to trivial learn the value in the user's data that failed validation.  

If it's not trivial to include `message` here, it's likely not worthwhile.

ingest/ingest_pipeline.py
```diff
@@ -427,6 +445,16 @@ def subsample(self):
                         if load_status != 0:
                             return load_status
                 else:
+                    # Caution: recording errorTypes in this manner can clobber other collected errors.
+                    # In subsampling, known failure modes are ValueErrors which stop processing so
+                    # this logging approach should not lose file validation information
+                    config.get_metric_properties().update(
+                        {
+                            "errorTypes": ["content:missing:values-across-files"],
+                            "numErrortypes": 1,
```
eweitz, line 454: Blocker: this trivial casing bug is worth fixing now, even if deferring the `set_mixpanel_num` fix described in my other comment:

```suggestion
                            "numErrorTypes": 1,
```

eweitz (changes requested):
I'm glad validation logging is improving!   

I note two trivial blockers, describe a related bug I just found, and inquire about slightly improving debuggability.

ingest/ingest_pipeline.py
jlchang: Hm, I had forgotten about `set_mixpanel_num` so the specification of numErrorTypes here is not necessary.
updated `set_mixpanel_num` in `45230a2`

Will create ticket for restoring error messages to mixpanel logging.

ingest/ingest_pipeline.py
jlchang: removed `"numErrortypes": 1`, should be handled by the fix in `set_mixpanel_num` in `45230a2`

jlchang:
Failed test unrelated to content of this PR
Failure in GenomesTestCase.test_genomes_default 
error was:  ConnectionResetError: [Errno 104] Connection reset by peer
marking this build failure as False positive due to 3rd party dependency

eweitz (approved):

bistline (approved):

PR #240, opened 2022-03-28
Update pip in docker build to resolve pandas install issue

jlchang:
The update of pandas in `jlc_make_data_objects` caused a cloud build failure because more recent pandas' setup.py invoke `import numpy`. In a cloud build scenario, numpy is collected but not installed when the pandas build is attempted so the pandas build fails (Older pip builds each package first before installing any packages). 
Screen Shot 2022-03-28 at 11 48 50 AM

By updating pip as part of the docker install process, a more sophisticated install solver uses better order of installation to ensure numpy is available before pandas installation is attempted.

Screen Shot 2022-03-28 at 12 48 22 PM

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,6 +23,7 @@ RUN apt-get -y update && \
   apt -y install python3.7-dev
 
 RUN python3.7 -m pip install pip
+RUN python3.7 -m pip install --upgrade pip
 
 # Set cleaner defaults (`alias` fails)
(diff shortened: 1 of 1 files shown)


devonbush (approved):

eweitz (approved):

jlchang:
Had CI failure for `test_genome` with `http.client.IncompleteRead: IncompleteRead(17437892 bytes read, 33391148 more expected)`
Retried CI with success - false positive was due to transient issues with a 3rd party dependency.

PR #242, opened 2022-03-30
Integration test for pilot DE process (SCP-4204)

jlchang:
Using a small test dataset derived from public study SCP1677, implemented an integration test for the pilot DE pipeline. This test runs the equivalent of:

> ingest_pipeline.py --study-id addedfeed000000000000000 --study-file-id dec0dedfeed0000000000000 differential_expression --annotation cell_type__ontology_label --matrix-file-path ../tests/data/differential_expression/de_integration.tsv --matrix-file-type dense --cell-metadata-file ../tests/data/differential_expression/de_integration_unordered_metadata.tsv --cluster-file ../tests/data/differential_expression/de_integration_cluster.tsv --name de_integration --study-accession SCPde --differential_expression 

and checks that expected output file "de_integration--cell_type__ontology_label--cholinergic_neuron--wilcoxon.tsv"
has the same md5 checksum as the checksum calculated using the reference file in `tests/data/differential_expression/reference`

To test locally, in the test directory, run:
pytest -k test_de_process
and confirm that it passes.

This satisfies SCP-4204.

### DIFF
--- a/ingest/de.py
+++ b/ingest/de.py
@@ -159,7 +159,7 @@ class DifferentialExpression:
                 self.barcodes,
             )
             DifferentialExpression.de_logger.info("preparing DE on sparse matrix")
-        else:
+        elif self.matrix_file_type == "dense":
             self.run_h5ad(
                 self.cluster,
                 self.metadata,
@@ -171,6 +171,12 @@ class DifferentialExpression:
                 self.method,
             )
             DifferentialExpression.de_logger.info("preparing DE on dense matrix")
+        else:
+            msg = f"Submitted matrix_file_type should be 'dense' or 'mtx' not {self.matrix_file_type}"
+            log_exception(
+                DifferentialExpression.dev_logger, DifferentialExpression.de_logger, msg
+            )
+            raise ValueError(msg)
 
     @staticmethod
     def run_h5ad(
--- a/ingest/de.py
+++ b/ingest/de.py
@@ -172,7 +172,7 @@ class DifferentialExpression:
             )
             DifferentialExpression.de_logger.info("preparing DE on dense matrix")
         else:
-            msg = f"Submitted matrix_file_type should be 'dense' or 'mtx' not {self.matrix_file_type}"
+            msg = f"Submitted matrix_file_type should be \"dense\" or \"mtx\" not {self.matrix_file_type}"
             log_exception(
                 DifferentialExpression.dev_logger, DifferentialExpression.de_logger, msg
             )
--- a/tests/test_de.py
+++ b/tests/test_de.py
@@ -5,6 +5,8 @@
 import unittest
 import sys
 import hashlib
+import os
+import pandas as pd
 
 sys.path.append("../ingest")
 from cell_metadata import CellMetadata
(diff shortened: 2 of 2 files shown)


ingest/de.py
```diff
@@ -171,6 +171,12 @@ def execute_de(self):
                 self.method,
             )
             DifferentialExpression.de_logger.info("preparing DE on dense matrix")
+        else:
+            msg = f"Submitted matrix_file_type should be 'dense' or 'mtx' not {self.matrix_file_type}"
```
eweitz, line 175:
```suggestion
            msg = f"Submitted matrix_file_type should be \"dense\" or \"mtx\" not \"{self.matrix_file_type}\""
```

eweitz (approved):

devonbush (changes requested):
As we talked about, let's add a check for the number of files generated, the filenames, and a spot check of some values

ingest/de.py
jlchang: `7205b9b`

jlchang:
Added more assertions to `test_de_process` to check presence of expected output files, file naming and spot-check of file content.

tests/test_de.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+def find_expected_files(labels, cluster_name, annotation, method):
+    """ Check that files were created for all expected annotation labels
+    """
+    found = 0
+    for label in labels:
+        sanitized_label = label.replace(" ", "_")
+        expected_file = (
+            f"{cluster_name}--{annotation}--{str(sanitized_label)}--{method}.tsv"
```
eweitz, line 36: `str` seems extraneous; `replace` assumes string, and f-string interpolation coerces to string.

```suggestion
        expected_file = (
            f"{cluster_name}--{annotation}--{sanitized_label}--{method}.tsv"
```

ingest/de.py
```diff
@@ -171,6 +171,12 @@ def execute_de(self):
                 self.method,
             )
             DifferentialExpression.de_logger.info("preparing DE on dense matrix")
+        else:
+            msg = f"Submitted matrix_file_type should be \"dense\" or \"mtx\" not {self.matrix_file_type}"
```
eweitz, line 175: Slightly easier readability:

```suggestion
            msg = f"Submitted matrix_file_type should be \"dense\" or \"mtx\" not \"{self.matrix_file_type}\""
```

eweitz (approved):
Looks good!  I noted some non-blocking nits.

tests/test_de.py
jlchang: Good point. The string coercion was put in place before I started sanitizing the label so any coercion is needed before the sanitizing step. So I need to step back and think about what is needed if a cell is not annotated with cell_type__ontology_label (cell_type is optional metadata so this is reasonable in our validated metadata). For those cases the label would be NaN which is a float (and would be represented in visualizations as "Unspecified").

devonbush (approved):
Nice, appreciate the spot-checks

tests/test_de.py
jlchang: I work on addressing the above issue in`jlc_sparse_de` to avoid more "artifactual merge commits"

PR #248, opened 2022-04-16
Setting __event_name for unhandled errors (SCP-4279)

devonbush:
Fixes a bug where `__event_name` was read in the `exit_pipeline` function, but not set if the global exception handler was triggered.

### DIFF
--- a/ingest/ingest_pipeline.py
+++ b/ingest/ingest_pipeline.py
@@ -600,6 +600,7 @@ def main() -> None:
             print(f'{key}: {metrics_dump[key]}')
 
     except Exception as e:
+        config.set_parent_event_name("ingest-pipeline:unhandled-exception:ingest")
         log_exception(IngestPipeline.dev_logger, IngestPipeline.user_logger, e)
         status = 1
 
--- a/ingest/ingest_pipeline.py
+++ b/ingest/ingest_pipeline.py
@@ -576,17 +576,22 @@ def main() -> None:
     Returns:
         None
     """
+    parsed_args = create_parser().parse_args()
+    validate_arguments(parsed_args)
+    arguments = vars(parsed_args)
     status = 0
+    status_cell_metadata = 0
+    ingest = None
+
+    # Initialize global variables for current ingest job
+    config.init(
+        arguments["study_id"],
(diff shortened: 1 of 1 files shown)


ingest/ingest_pipeline.py
```diff
@@ -576,29 +576,35 @@ def main() -> None:
...
     """
+    parsed_args = create_parser().parse_args()
+    validate_arguments(parsed_args)
+    arguments = vars(parsed_args)
     status = 0
+    status_cell_metadata = 0
+    ingest = None
+
+    # Initialize global variables for current ingest job
+    config.init(
```
jlchang, line 587: local testing shows this erroring out:
```
Traceback (most recent call last):  
  File "ingest_pipeline.py", line 614, in <module>  
    main()  
  File "ingest_pipeline.py", line 608, in main  
    MetricsService.log(config.get_parent_event_name(), config.get_metric_properties())  
  File "/Users/jlchang/Documents/GitHub/scp-ingest-pipeline/ingest/config.py", line 28, in get_parent_event_name  
    return __event_name  
NameError: name '__event_name' is not defined
```
Moving `try:` down one line so the IngestPipeline object is instantiated should resolve this error.

jlchang (changes requested):
Having `try:` above IngestPipeline object instantiation results in `NameError: name '__event_name' is not defined` because config isn't fully initialized ([more details](https://broadinstitute.slack.com/archives/CBEHTH601/p1650115317813179?thread_ts=1650107131.342889&cid=CBEHTH601))

Shifting the `try:` statement down one line should resolve the NameError.

Not entirely clear if it will yield the logging (Sentry or log files) we're hoping for, but it should get us closer...

eweitz (approved):
Could you add a PR description, for some context?  

To my understanding, the intent here is to improve observability of unhandled errors in Mixpanel -- something like that seems like a more useful PR title, to help us recall what this PR did after it's no longer top-of-mind.

That said, code looks good!

jlchang (approved):
Confirmed that simulated error will result in stack trace to `log.txt`
confirmed Mixpanel logging of `ingest-pipeline:unhandled-exception:ingest` events

PR #270, opened 2022-09-06
Parallelizing expression matrix processing for Image Pipeline (SCP-4647)

bistline:
#### BACKGROUND
Building on the work from SCP-4584, it will be necessary to optimize performance of jobs that process expression data so that they can complete as quickly as possible to allow Image Pipeline to create images faster.  This will reduce the lag between a user uploading an expression matrix and each gene-level entry having a preview image available for viewing.

#### CHANGES
This update introduces several changes.  The largest change is that files are now "chunked" into discrete parts, where the size of each chunk is the total file size divided by the number of available cores on the machine (minus 1).  Each chunk has logic to ensure that it terminates on a newline so as to avoid incomplete data.  Secondly, sparse matrix files now no longer need to be sorted on gene indices before being processed, as the chunks are now processed out into individual gene-level sparse matrix fragments, which each worker process appending to the fragments.  Since these are all single-gene level entries, it does not matter what order the lines are in as the indexes for barcodes remain unchanged.  In addition, missing genes from the sparse matrix (i.e. no observed expression for a given gene from the genes file) will be rendered out as arrays of `0` for each cell in the cluster file.  Lastly, the script `render_expression_arrays.py` has been refactored into the class `ExpressionWriter`, which will allow for easier integration with the rest of the codebase.

Running on a 64 cores GCP machine, this updated script was able to process the 1.3M cell 10X benchmark dataset in 74 minutes (down from ~14 hours for the single-process version).

#### MANUAL TESTING
1. Pull branch and navigate to `scripts/scratch_ingest`
2. Run the following command to process a dense matrix:
```
python3 expression_writer.py --matrix-file ../../tests/data/dense_expression_matrix.txt \ 
                             --cluster-file ../../tests/data/cluster_example.txt \
                             --cluster-name 'Dense Example'
```
3. Confirm that you see messages detailing the chunking logic (note the last chunk has 0 bytes, this is due to the small size of the file and has no impact on output):
```
determining seek points for ../../tests/data/dense_expression_matrix.txt with chunk size 37
reading ../../tests/data/dense_expression_matrix.txt at index 161:264
reading ../../tests/data/dense_expression_matrix.txt at index 265:265
```
4. Confirm that two gene-level JSON arrays are generated - `Dense_Example--Itm2a.json.gz` and `Dense_Example--Sergef.json.gz`
5. Open both files, and confirm that `Itm2a` has all zeros, and `Sergef` has 8 non-zero entries
6. Run the following command to process a sparse matrix:
```
python3 expression_writer.py --matrix-file ../../tests/data/mtx/matrix_with_header.mtx \
                             --genes-file ../../tests/data/mtx/sampled_genes.tsv \
                             --barcodes-file ../../tests/data/mtx/barcodes.tsv \
                             --cluster-file ../../tests/data/mtx/cluster_mtx_barcodes.tsv \
                             --cluster-name 'Sparse Example'
```
7. Confirm chunking statements:
```
determining seek points for ../../tests/data/mtx/matrix_with_header.mtx with chunk size 31
reading ../../tests/data/mtx/matrix_with_header.mtx at index 56:94
reading ../../tests/data/mtx/matrix_with_header.mtx at index 95:135
reading ../../tests/data/mtx/matrix_with_header.mtx at index 136:177
reading ../../tests/data/mtx/matrix_with_header.mtx at index 178:219
reading ../../tests/data/mtx/matrix_with_header.mtx at index 220:220
```
8. In the output directory, confirm there are 7 files, as well as 6 gene-level sparse matrix files in the `gene_entries` folder (there will not be an entry for `HOMER2`)
9. Open `Sparse_Example--HOMER2.json.gz` and confirm it has only zeros, and any of the other files have 2 non-zero entries.
10. (Optional) Re-run any of the above commands with `--precision 1` and confirm that output files have only 1 digit of precision for non-zero values


scripts/scratch_ingest/expression_writer.py
```diff
@@ -0,0 +1,416 @@
...
+        :param fragment_name: (String) name of gene-level fragment file
+        """
+        gene_name = fragment_name.split('__')[0]
+        full_path = f"{data_dir}/gene_entries/{fragment_name}"
+        observed_cells = []
+        exp_vals = []
+        # write array of 0 values if file is empty
+        if os.stat(full_path).st_size == 0:
+            empty_exp = [0] * len(cluster_cells)
+            ExpressionWriter.write_gene_scores(cluster_name, gene_name, empty_exp, data_dir)
```
eweitz: This is another area ripe for optimization, likely worth noting as a TODO.

You mentioned roughly 4k of 28k genes had no expression in our pilot dataset.  So we could write a list of gene symbols that lack expression in a special file, then background-read that into the frontend upon loading a cluster, and display a “0” gene expression scatter plot for all of them.  Then we could skip writing these thousands of files of all zeros.

That'd save sizable pipeline run time, end-user / Image Pipeline download and parse time, and storage cost.

scripts/scratch_ingest/expression_writer.py
```diff
@@ -0,0 +1,416 @@
...
+        :param cluster_cells: (List) cluster cell names
+        :param exp_cells: (List) expression cell names
+        :param exp_scores: (List) expression values, in the same order as exp_cells
+        :return: (List) Expression values, ordered by cluster_cells
+        """
+        observed_exp = dict(zip(exp_cells, exp_scores))
+        return (observed_exp.get(cell, 0) for cell in cluster_cells)
+
+    @staticmethod
+    def write_gene_scores(cluster_name, gene_name, exp_values, data_dir):
```
eweitz: Optimization idea: upload small batches of files sooner instead of all files last.  This method seems relevant there.

To start, I imagine we'd write all expression array files to disk, then upload (`gsutil -m cp...`) them all to the bucket.  That approach would make any per-gene downstream pipelines (e.g. Image Pipeline) wait until the scatter data pipeline totally completes.  To allow downstream pipelines to start sooner, this pipeline could -- in a background thread -- upload expression array files in sets of, say, 100, to the bucket, while updating a file listing uploaded genes.  (This `write_gene_scores` method would presumably increment a related counter.)  That would let downstream pipelines pull the gene list index, then process expression arrays as they become available.

This optimization would decrease image cache warming times likely by > 1 hour for this pilot case.

If that sounds feasible to you, I'd be happy to create a ticket, then we can further assess effort-value tradeoffs and priority relative to other optimizations.

scripts/scratch_ingest/expression_writer.py
```diff
@@ -0,0 +1,416 @@
...
+        :param matrix_cells: (List) cells from header line of dense matrix
+        :param cluster_cells: (List) cell names from cluster file
+        :param cluster_name: (String) name of cluster object
+        :param data_dir (String) name out output dir
+        :param line: (String) single line from dense matrix
+        """
+        clean_line = line.rstrip().replace('"','')
+        raw_vals = re.split(COMMA_OR_TAB, clean_line)
+        gene_name = raw_vals.pop(0)
+        exp_vals = [round(float(val), self.precision) if float(val) != 0.0 else 0 for val in raw_vals]
```
eweitz, line 351: Per other comment re `round_exp`:
```suggestion
        exp_vals = [round_exp(val, self.precision) if float(val) != 0.0 else 0 for val in raw_vals]
```

scripts/scratch_ingest/expression_writer.py
```diff
@@ -0,0 +1,435 @@
...
+        """
+        Process a single line from a sparse matrix and extract values as integers
+
+        :param line: (String) single line from matrix file
+        :return: (List): values as integers
+        """
+        gene_idx, barcode_idx, raw_exp = line.rstrip().split(' ')
+        return [int(gene_idx), int(barcode_idx), round(float(raw_exp), self.precision)]
+
+    def write_empty_sparse_genes(self, genes, num_cluster_cells, cluster_name, data_dir):
```
eweitz, line 281: Optimization idea maybe worth a TODO: 

Write a list of gene symbols that lack expression in a special file, then background-read that into the frontend upon loading a cluster, and display a “0” gene expression scatter plot for all of them.  

Given ~4k of 28k genes were empty in our pilot, that would save a sizable amount of total data size and run time here, as well as time for end users and Image Pipeline.

eweitz (approved):
Prototype code looks good!  I suggest optimization and maintainability refinements, no blockers.

scripts/scratch_ingest/expression_writer.py
```diff
@@ -0,0 +1,416 @@
...
+        os.mkdir(dirname)
+        os.mkdir(f"{dirname}/gene_entries") # for slicing sparse matrix files
+        return dirname
+
+    @staticmethod
+    def get_matrix_size(matrix_file_path):
+        """
+        Get the actual size of a matrix file, whether gzipped or not
+        Because gzipped files only use 4 bytes to store the 'original' file size,
+        any files larger that 4GB will only store the modulus of 2 ** 32, not the
```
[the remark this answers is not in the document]
bistline: Always mess that up...

scripts/scratch_ingest/expression_writer.py
```diff
@@ -0,0 +1,416 @@
...
+        """
+        Write a JSON array of expression values
+        Filename uses {cluster_name}--{gene_name}.json format
+
+        :param cluster_name: (String) Name of cluster
+        :param gene_name: (String) Name of gene
+        :param exp_values: (List) expression values
+        :param data_dir: (String) name out output dir
+        """
+        with gzip.open(f"{data_dir}/{cluster_name}--{gene_name}.json.gz", "wt") as file:
```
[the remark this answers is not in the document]
bistline: I'm not entirely sure that it matters - I don't see anything about compression levels in their docs on [transcoding](https://cloud.google.com/storage/docs/transcoding) - so it could be that the answer is "yes".  We can always try and see what happens.

I did find that `gsutil cp` has a `-Z` flag which will _automatically gzip all files on upload_, so we may in fact want to write non-compressed and let `gsutil` handle it for us, as this will also set the appropriate `Content-Encoding` and `Cache-Control` headers.

scripts/scratch_ingest/expression_writer.py
bistline: I'm a little less enthused about this one, though on its face it makes sense.  I'm just worried about concurrency/polling.  It is much easier to architect as successive jobs being dispatched from the portal, instead of two concurrent jobs.  That being said it's definitely still worth ticketing and we can prioritize as we see fit.

scripts/scratch_ingest/expression_writer.py
bistline: I think that's definitely worth ticketing, but for now it will keep the code on your end cleaner.

bistline:
> A general non-blocking maintainability suggestion: I'd generally favor standalone functions over static methods. That practice has worked well in SCP JavaScript. Functions are flatter and seem easier to reuse.

Once I started looking at that, it makes complete sense - I've refactored to add in a separate library with all the static methods.

scripts/scratch_ingest/expression_writer.py
```diff
@@ -0,0 +1,419 @@
...
+    @staticmethod
+    def is_gz_file(filepath):
+        """
+        Determine if a file is gzipped by checking first two bytes against GZIP_MAGIC_NUM
+
+        :param filepath: (String) path to file to check
+        :returns (Bool)
+        """
+        with open(filepath, 'rb') as test_f:
+            return test_f.read(2) == GZIP_MAGIC_NUM
```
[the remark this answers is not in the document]
bistline: [36e7b90](https://github.com/broadinstitute/scp-ingest-pipeline/pull/270/commits/36e7b906849129a2177b865d7705ef033995fdbe)

scripts/scratch_ingest/expression_writer.py
```diff
@@ -0,0 +1,419 @@
...
+
+    @staticmethod
+    def open_file(filepath):
+        """
+        Open a file with the correct reader
```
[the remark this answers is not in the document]
bistline: [36e7b90](https://github.com/broadinstitute/scp-ingest-pipeline/pull/270/commits/36e7b906849129a2177b865d7705ef033995fdbe)

scripts/scratch_ingest/expression_writer.py
```diff
@@ -0,0 +1,419 @@
...
+    ):
+        self.matrix_file_path = matrix_file_path
+        self.cluster_file_path = cluster_file_path
+        self.cluster_name = cluster_name
+        self.genes_file = genes_file
+        self.barcodes_file = barcodes_file
+        if num_digits is not None:
+            self.precision = num_digits
+        if cores is not None:
+            self.num_cores = cores
```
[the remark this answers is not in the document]
bistline: [36e7b90](https://github.com/broadinstitute/scp-ingest-pipeline/pull/270/commits/36e7b906849129a2177b865d7705ef033995fdbe)

scripts/scratch_ingest/expression_writer.py
bistline: [36e7b90](https://github.com/broadinstitute/scp-ingest-pipeline/pull/270/commits/36e7b906849129a2177b865d7705ef033995fdbe)

scripts/scratch_ingest/expression_writer.py
```diff
@@ -0,0 +1,210 @@
...
+        """
+        cluster_file_path = args.cluster_file
+        expression_file_path = args.matrix_file
+        sanitized_cluster_name = re.sub(r'\W', '_', args.cluster_name)
+        data_dir = make_data_dir(sanitized_cluster_name)
+        cluster_cells = get_cluster_cells(cluster_file_path)
+        if self.genes_file is not None and self.barcodes_file is not None:
+            print(f"reading {expression_file_path} as sparse matrix")
+            genes_file = open_file(args.genes_file)
+            genes = load_entities_as_list(genes_file, 1)
```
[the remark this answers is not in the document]
bistline: Ah yes - excellent catch!  That should be relatively straightforward.

bistline:
> I'm excited to think about how this approach might be incorporated into ingest pipeline for expression matrix ingest. Could we generate these files once and use them for both Image Pipeline _and_ expression ingest?

That is my supposition.  At the very least, we could apply this paradigm to sparse matrices and completely remove the need to sort them ahead of time before doing the actual reads/inserts into the database.  And if we combined the two together - ingest and generating assets for image pipeline - we could completely remove any lag between the two.

scripts/scratch_ingest/expression_writer.py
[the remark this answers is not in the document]
bistline: [a4e68b3](https://github.com/broadinstitute/scp-ingest-pipeline/pull/270/commits/a4e68b3fd5dfdec08db97c41cbc369f4333871a1)

jlchang (approved):
Code looks good!

scripts/scratch_ingest/expression_writer.py
eweitz: I just opened SCP-4683: "Optimize scatter data speed/cost 14% via unexpressed genes".

PR #271, opened 2022-09-21
Handle existing inserts when retrying from dropped MongoDB connection (SCP-4621)

bistline:
#### BACKGROUND
Very large expression matrices (e.g. > 10 GiB) are prone to connection issues during ingest.  While the cause is not fully understood, one theory is that expression matrices with both high cell counts & high proportions of expressed cells cause timeouts when inserting into MongoDB due to the large amount of data sent to and received back from the database.  Current retry functionality has been shown to be very brittle with extremely large files, causing repeated failed ingests.

#### CHANGES
While investigating a 47 GiB matrix that failed to ingest, it was discovered that our logic for filtering out documents that may have already been inserted before the connection was dropped was faulty, and as such may not catch all potential errors.  The updated filtering methodology only matches on the document `name` and `array_index` (rather than the entire document) as these two properties alone will constitute a unique key.  This update also filters out the `values` array from `DataArray` documents when logging errors to make debugging easier for extremely large arrays.

This update also corrects an issue with building the Docker image locally where `apt` cannot connect to some repositories and is replaced with `apt-get`, which does not appear to have the same issues.

#### MANUAL TESTING
Since this problem only manifests when attempting to ingest very large matrices, manual testing cannot be done in the context of a normal PR review (aside from running the tests) as the ingest process takes hours to complete.  However, this updated code has ingested the entire 47 GiB dense matrix twice, taking slightly over 4 hours both times.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,11 +16,11 @@ FROM marketplace.gcr.io/google/ubuntu1804:latest
 
 # Install Python 3.7
 RUN apt-get -y update && \
-  apt -y install software-properties-common && \
+  apt-get -y install software-properties-common && \
   add-apt-repository ppa:deadsnakes/ppa && \
-  apt -y install python3-pip && \
-  apt -y install python3.7 && \
-  apt -y install python3.7-dev
+  apt-get -y install python3-pip && \
+  apt-get -y install python3.7 && \
+  apt-get -y install python3.7-dev
 
 RUN python3.7 -m pip install --upgrade pip
 
--- a/ingest/mongo_connection.py
+++ b/ingest/mongo_connection.py
@@ -90,18 +90,20 @@ def graceful_auto_reconnect(mongo_op_func):
                     else:
                         return args[0]
                 else:
-                    filtered_bwe = filter_values_from_log(bwe)
-                    dev_logger.debug(str(filtered_bwe.details))
-                    raise filtered_bwe
+                    log_error_without_values(bwe)
+                    raise bwe
 
     return wrapper
 
-def filter_values_from_log(error):
-    error_doc = error.details["writeErrors"][0]['op']
-    if 'values' in error_doc:
-        error_doc['values'] = '<filtered>'
-    error.details["writeErrors"][0]['op'] = error_doc
-    return error
+def log_error_without_values(error):
+    """Remove 'values' array from log messages as this is usually very long and makes logs difficult to read
+
+    :param error: BulkWriteError from failed transaction
+    """
+    for entry in error.details["writeErrors"]:
+        if 'values' in entry['op']:
+            entry['op']['values'] = '<filtered>'
+    dev_logger.debug(str(error.details))
 
 def discard_inserted_documents(errors, original_documents):
     """Discard any documents that have already been inserted which are violating index constraints
--- a/ingest/mongo_connection.py
+++ b/ingest/mongo_connection.py
@@ -117,16 +117,14 @@ def discard_inserted_documents(error_doc, original_documents, collection_name, m
            List[Dict]: list of documents with existing entries removed
     """
     names_to_find = list(doc['name'] for doc in original_documents)
-    study_id = error_doc['study_id']
-    study_file_id = error_doc['study_file_id']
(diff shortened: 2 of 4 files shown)


bistline:
Ingest failed after update to account for `array_index`, so more work is needed.

ingest/mongo_connection.py
```diff
@@ -76,40 +76,51 @@ def wrapper(*args, **kwargs):
...
        such documents will have an error code of 11000 for a DuplicateKey error
        from https://github.com/mongodb/mongo/blob/master/src/mongo/base/error_codes.yml#L467
+       Uses Mongo query to get existing names and filters against documents to be inserted as
+       only first error is returned, not all possible errors
 
        Parameters:
-           error_documents (List[Dict]): list of documents that failed to insert in original transaction
-           original_documents (List[Dict]): list of documents from original transaction that failed
-           error_code (Int): error status code to filter on
+           errors (List<Dict>): Documents that failed to insert from transaction
```
eweitz, line 114: Closer to Python convention for documenting types, i.e. closer to [type hints](https://docs.python.org/3/library/typing.html):

```suggestion
           errors (List[Dict]): Documents that failed to insert from transaction
```

ingest/mongo_connection.py
```diff
@@ -76,40 +76,51 @@ def wrapper(*args, **kwargs):
...
+           original_documents (List<Dict>): list of documents from original transaction that failed
 
        Returns:
-           List[Dict]: list of documents with matching error code entries removed
+           List<Dict>: list of documents with existing entries removed
```
eweitz, line 118: Declaring [a list of dicts as the return type](https://stackoverflow.com/questions/68199174/python-typing-listdict-vs-listdict) would be more idiomatic than this, but if we keep it in a docstring for now we should at least use Python-conventional square brackets:

```suggestion
           List[Dict]: list of documents with existing entries removed
```

Dockerfile
```diff
@@ -16,11 +16,11 @@ FROM marketplace.gcr.io/google/ubuntu1804:latest
 
 # Install Python 3.7
 RUN apt-get -y update && \
-  apt -y install software-properties-common && \
+  apt-get -y install software-properties-common && \
```
eweitz, line 19: Thanks for explaining this in the PR description.

ehanna4 (approved):
Looks good! Didn't run the tests but looking them over was helpful to understand the change more and it looks okay to me👍

eweitz (approved):
Code looks good!  I note two non-blocking nits.

Kudos on the promising fix -- this is a tough problem.  Given the fix has no feasible automated or manual test, here are a few questions to assess our confidence in it:
1.  Did the 47 GiB file fail on all attempts before?
2. How many attempts failed, roughly?

My vague recollection is the 47 GiB file failed on every of > 5 ingest attempts, and failed in a fairly consistent manner, so the fact that this succeeds should instill relatively high confidence.  But you have more context!

bistline:
Great questions!  To answer:

> * Did the 47 GiB file fail on all attempts before?
Yes - it failed every time the user uploaded the file, and on 4 attempts locally as I was investigating fixes.  The first successful ingest was yesterday, but I realized that the filtering functionality wasn't taking into account `array_index` on `DataArray` documents, meaning that we could potentially be dropping data.  Once I addressed that, I got it working again and had another successful ingest that finished at ~3:15pm today.  I'm running it through ingest again with the same image to confirm that it still works.
> * How many attempts failed, roughly?
7 total - 4 of mine, 3 from the user.

ingest/mongo_connection.py
bistline: Ahh good to know - can do.

bistline: [14532dc](https://github.com/broadinstitute/scp-ingest-pipeline/pull/271/commits/14532dcd67b2b4aecc1e19c923883cef6352c913)

ingest/mongo_connection.py
bistline: [14532dc](https://github.com/broadinstitute/scp-ingest-pipeline/pull/271/commits/14532dcd67b2b4aecc1e19c923883cef6352c913)

PR #273, opened 2022-10-03
Integrating `ExpressionWriter` into `IngestPipeline` (SCP-4648)

bistline:
#### BACKGROUND
In order for Image Pipeline to have the necessary pre-rendered expression artifacts available with which to produce static expression scatter plot images, the `ExpressionWriter` class needs to be integrated into the rest of `IngestPipeline` so that it can be launched as a PAPI job.  This means that the CLI needs to be invoked through `ingest_pipeline.py`, and that it can localize/delocalize files from GCP buckets, and have adequate test coverage as a part of normal CI for `scp-ingest-pipeline`.

#### CHANGES
This update changes `ExpressionWriter` from a standalone class to one that can be invoked through `IngestPipeline`.  Basic functionality is unchanged, though file handling is now delegated to the `IngestFiles` class, which can push/pull files directly from GCP buckets.  In addition, logging is now handled by the `monitor.py` module, and MixPanel/Sentry integration is handled through normal codepaths in `IngestPipeline`.   A custom delocalization method was added to handle the expected ~25-30K file export at the end of a successful run.  This will leverage `gsutil` in the host OS to allow for parallel file uploads, as the Python `google.cloud.storage` package does not support this (only serial uploads, and it specifically cannot be parallelized via `multiprocessing`).

#### MANUAL TESTING
1. Pull this branch and initialize the `scp-ingest-pipeline` environment:
```
python3 -m venv env --copies
source env/bin/activate
pip install -r requirements.txt
```
2. Setup MongoDB integration with the following, passing the path to your vault token (we don't need to talk to the database, but ingest will fail to execute without it):
```
source scripts/setup_mongo_dev.sh ~/path/to/github-token-for-vault.txt
```
3. In a separate terminal, get the bucket ID and browser link for any study with the following:
```
study = Study.all.sample # you can use a specific study if you want, it doesn't matter
bucket_id = study.bucket_id
study.google_bucket_url
``` 
4. Copy the URL from above and open it in a browser window, and then copy the following files to the bucket:
  * `tests/data/dense_expression_matrix.txt`
  * `tests/data/cluster_example.txt`
  * `tests/data/mtx/matrix_with_header.mtx`
  * `tests/data/mtx/cluster_mtx_barcodes.tsv`
  * `tests/data/mtx/sampled_genes.tsv`
  * `tests/data/mtx/barcodes.tsv`
5. Using the bucket ID from above, run a job using the dense matrix (the `study_id` and `study_file_id` values aren't used right now, so don't worry about getting valid ones):
```
BUCKET_ID="fc-c36727b0-1663-40fc-ac37-9016a6709829"
python3 ingest_pipeline.py --study-id 5d276a50421aa9117c982845 --study-file-id 5dd5ae25421aa910a723a337 render_expression_arrays \
                             --matrix-file-path gs://$BUCKET_ID/dense_expression_matrix.txt \
                             --matrix-file-type dense \
                             --cluster-file gs://$BUCKET_ID/cluster_example.txt \
                             --cluster-name 'Dense Example' --render-expression-arrays
```
6. You should see output similar to this:
```
studyAccession: SCPdev
fileName: 5dd5ae25421aa910a723a337
fileSize: 1
trigger: dev-mode
logger: ingest-pipeline
appId: single-cell-portal
(env) bistline@wm306-020 ingest % If you experience problems with multiprocessing on MacOS, they might be related to https://bugs.python.org/issue33725. You can disable multiprocessing by editing your .boto config or by adding the following flag to your command: `-o "GSUtil:parallel_process_count=1"`. Note that multithreading is still available even if you disable multiprocessing.

Copying file://Dense_Example/Setd6.json.gz [Content-Type=application/json]...
Copying file://Dense_Example/Gm12000.json.gz [Content-Type=application/json]... 
Copying file://Dense_Example/Sergef.json.gz [Content-Type=application/json]...  
Copying file://Dense_Example/Itm2a.json.gz [Content-Type=application/json]...   
Copying file://Dense_Example/1700061E18Rik.json.gz [Content-Type=application/json]...
Copying file://Dense_Example/Fam71e2.json.gz [Content-Type=application/json]... 
Copying file://Dense_Example/Chil5.json.gz [Content-Type=application/json]...   
Copying file://Dense_Example/Gstm2.json.gz [Content-Type=application/json]...   
Copying file://Dense_Example/Fam109a.json.gz [Content-Type=application/json]... 
Copying file://Dense_Example/Gm14444.json.gz [Content-Type=application/json]... 
Copying file://Dense_Example/Vps28.json.gz [Content-Type=application/json]...
Copying file://Dense_Example/Hebp2.json.gz [Content-Type=application/json]...   
Copying file://Dense_Example/Psma4.json.gz [Content-Type=application/json]...   
Copying file://Dense_Example/Mks1.json.gz [Content-Type=application/json]...    
Copying file://Dense_Example/Ssu72.json.gz [Content-Type=application/json]...   
Copying file://Dense_Example/Eif2b2.json.gz [Content-Type=application/json]...  
Copying file://Dense_Example/Olfr1018.json.gz [Content-Type=application/json]...
Copying file://Dense_Example/Spn-ps.json.gz [Content-Type=application/json]...  
Copying file://Dense_Example/Dhx9.json.gz [Content-Type=application/json]...    
/ [19/19 files][  1.1 KiB/  1.1 KiB] 100% Done                                  
Operation completed over 19 objects/1.1 KiB.                                     
```
7. In the browser window with the bucket, refresh the page and navigate to `_scp_internal/cache/expression_scatter/data/Dense_Example`
8. Confirm there are two files here: `Itm2a.json.gz` and `Sergef.json.gz`
9. Now run the sparse example:
```
BUCKET_ID="fc-c36727b0-1663-40fc-ac37-9016a6709829"
python3 ingest_pipeline.py --study-id 5d276a50421aa9117c982845 --study-file-id 5dd5ae25421aa910a723a337 \
                             render_expression_arrays --matrix-file-path gs://$BUCKET_ID/matrix_with_header.mtx \
                             --matrix-file-type mtx \
                             --gene-file gs://$BUCKET_ID/sampled_genes.tsv \
                             --barcode-file gs://$BUCKET_ID/barcodes.tsv \
                             --cluster-file gs://$BUCKET_ID/cluster_mtx_barcodes.tsv \
                             --cluster-name 'Sparse Example' --render-expression-arrays
```
10. Confirm similar output to above, and back in the bucket view, validate there are 7 new files for the following genes: `C1orf159`, `RP11-345P4.9`, `GABRD`, `THAP3`, `DNAJC11`, `OXCT2`, `HOMER2`


ingest/expression_writer.py
```diff
@@ -0,0 +1,268 @@
...
+        # localize main files, if needed
+        self.local_matrix_path = open_file(self.matrix_file_path)[1]
+        self.local_cluster_path = open_file(self.cluster_file_path)[1]
+
+        # set storage bucket name, if needed
+        self.bucket = self.get_storage_bucket_name()
+
+        timestamp = datetime.datetime.now().isoformat(sep="T", timespec="seconds")
+        url_safe_timestamp = re.sub(':', '', timestamp)
+        log_name = f"expression_scatter_images_{url_safe_timestamp}_log.txt"
```
eweitz, line 71: Blocker: to avoid confusion with the downstream log file:

```suggestion
        log_name = f"expression_scatter_data_{url_safe_timestamp}_log.txt"
```

ingest/expression_writer.py
```diff
@@ -0,0 +1,268 @@
...
+        get_cluster_cells, load_entities_as_list, process_sparse_fragment, write_gene_scores, process_dense_line, \
+        COMMA_OR_TAB
+    from monitor import setup_logger
+    from ingest_files import IngestFiles
+except ImportError:
+    from .writer_functions import round_exp, encode_cluster_name, open_file, make_data_dir, get_matrix_size, \
+        get_cluster_cells, load_entities_as_list, process_sparse_fragment, write_gene_scores, process_dense_line, \
+        COMMA_OR_TAB
+    from .monitor import setup_logger
+    from .ingest_files import IngestFiles
```
eweitz, line 45: I introduced this inelegant import technique to deal with a quirk of supporting CSFV in our public CLI / `manage_study.py`.  We no longer support the public CLI, so it'd make sense to remove these `try`/ `except` clauses in general.  And (in particular) we wouldn't call this module from the public CLI, so we can certainly omit it here.

```suggestion
from writer_functions import round_exp, encode_cluster_name, open_file, make_data_dir, get_matrix_size, \
    get_cluster_cells, load_entities_as_list, process_sparse_fragment, write_gene_scores, process_dense_line, \
    COMMA_OR_TAB
from monitor import setup_logger
from ingest_files import IngestFiles
```

ingest/expression_writer.py
```diff
@@ -0,0 +1,268 @@
...
+
+class ExpressionWriter:
+    DELOCALIZE_FOLDER = "_scp_internal/cache/expression_scatter/data"
```
eweitz, line 48: Ending folder paths with "/" has always struck me as clearer.  I do this in my code, but we should probably land on some consensus.

```suggestion
    DELOCALIZE_FOLDER = "_scp_internal/cache/expression_scatter/data/"
```

This is my subjective preference; ideally I'd examine how such strings are ended more broadly.

ingest/expression_writer.py
```diff
@@ -0,0 +1,268 @@
...
+        self.delocalize_outputs(sanitized_cluster_name)
+
+    def delocalize_outputs(self, cluster_name):
+        """
+        Copy all output files to study bucket in parallel using gsutil (since there are usually ~25-30K files)
+
+        :param cluster_name: (str) encoded name of cluster
+        """
+        if self.bucket is not None:
+            bucket_path = f"{self.bucket}/{self.DELOCALIZE_FOLDER}/{cluster_name}"
```
eweitz, line 262:
```suggestion
            bucket_path = f"{self.bucket}/{self.DELOCALIZE_FOLDER}{cluster_name}"
```

ingest/expression_writer.py
```diff
@@ -0,0 +1,268 @@
...
+    from .writer_functions import round_exp, encode_cluster_name, open_file, make_data_dir, get_matrix_size, \
+        get_cluster_cells, load_entities_as_list, process_sparse_fragment, write_gene_scores, process_dense_line, \
+        COMMA_OR_TAB
+    from .monitor import setup_logger
+    from .ingest_files import IngestFiles
+
+class ExpressionWriter:
+    DELOCALIZE_FOLDER = "_scp_internal/cache/expression_scatter/data"
+    denominator = 2 if re.match('darwin', sys.platform) else 1
+    num_cores = int(multiprocessing.cpu_count() / denominator) - 1
```
eweitz, line 50: An edge case (running on a 1-CPU machine) that wouldn't hurt to handle:

```suggestion
    num_cores_available = int(multiprocessing.cpu_count() / denominator) - 1
    num_cores = num_cores_available if num_cores_available > 0 else 1
```

ingest/writer_functions.py
```diff
@@ -0,0 +1,203 @@
...
+
+def is_gz_file(filepath) -> bool:
+    """
+    Determine if a file is gzipped by checking first two bytes against GZIP_MAGIC_NUM
+
+    :param filepath: (str) path to file to check
+    :returns: (bool)
+    """
+    with open(filepath, 'rb') as test_f:
+        return test_f.read(2) == GZIP_MAGIC_NUM
```
eweitz, line 32: Let's use the [same exact name as core JS and Ruby](https://github.com/broadinstitute/single_cell_portal_core/search?q=gzip+magic+number), for good measure.

Supplemental notes:
* The fact that [JS only checks 1 byte](https://github.com/broadinstitute/single_cell_portal_core/blob/762ebd8fe9824a228e5133f99ef856e29877238c/app/javascript/lib/validation/validate-file-content.js#L323) and Ruby and this Python check 2 bytes makes me wary.
* Can this function replace the [equivalent function in ingest_files.py](https://github.com/broadinstitute/scp-ingest-pipeline/blob/913bf2b3f11c104bbbcddce0be4d2f6ed6be8859/ingest/ingest_files.py#L103)?

```suggestion
# The signature first byte of file compressed via gzip
GZIP_MAGIC_NUMBER = b'\x1f\x8b'

ALLOWED_FILE_TYPES = ["text/csv", "text/plain", "text/tab-separated-values"]

def is_gz_file(filepath) -> bool:
    """
    Determine if a file is gzipped by checking first two bytes against GZIP_MAGIC_NUMBER

    :param filepath: (str) path to file to check
    :returns: (bool)
    """
    with open(filepath, 'rb') as test_f:
        return test_f.read(2) == GZIP_MAGIC_NUMBER
```

ingest/writer_functions.py
```diff
@@ -0,0 +1,203 @@
...
+
+def round_exp(val, precision) -> float:
+    """
+    Round a raw value to a specified precision
+
+    :param val: (str) raw value to convert and round
+    :param precision: (int) number of digits of precision
+    :returns: (float)
```
eweitz, line 40: Nice use of type hints!  That makes this line extraneous.

```suggestion
```

ingest/writer_functions.py
```diff
@@ -0,0 +1,203 @@
...
+    """
+    return round(float(val), precision)
+
+def encode_cluster_name(name) -> str:
+    """
+    Encodes a cluster name to be used in a HTTP request
+    Will encode + signs as 'pos' to differentiate + and -
+    :param name: (str) name of cluster
+    :returns: (str) encoded cluster name
```
eweitz, line 49: Having the return type redundantly noted in a docstring means it could get out-of-sync -- a mild code smell.

```suggestion
    :returns: encoded cluster name
```

ingest/writer_functions.py
```diff
@@ -0,0 +1,203 @@
...
+def get_matrix_size(matrix_file_path) -> int:
+    """
+    Get the actual size of a matrix file, whether gzipped or not
+    Because gzipped files only use 4 bytes to store the 'original' file size,
+    any files larger that 4 GiB will only store the modulus of 2 ** 32, not the
+    actual size of the file.
+    Using seek() and tell() will return the last byte position
+
+    :param matrix_file_path: (str) path to matrix file
+    :returns: (int)
```
eweitz, line 83:
```suggestion
```

ingest/writer_functions.py
```diff
@@ -0,0 +1,203 @@
...
+        return matrix_file.tell()
+    else:
+        return os.stat(local_path).st_size
+
+def get_cluster_cells(file_path) -> list:
+    """
+    Return a list of cell names from a cluster file (1st column, starting line 3)
+
+    :param file_path: (str) absolute path to cluster_file
+    :returns: (list)
```
eweitz, line 97:
```suggestion
```

ingest/writer_functions.py
```diff
@@ -0,0 +1,203 @@
...
+            cells.append(cell)
+    return cells
+
+def load_entities_as_list(file) -> list:
+    """
+    Read an entire 10X feature/barcode file into a list for parsing sparse data
+
+    :param file: (TextIO) open file object
+    :param column: (int) specific column to extract from entity file
+    :returns: (list)
```
eweitz, line 114:
```suggestion
```

ingest/writer_functions.py
```diff
@@ -0,0 +1,203 @@
...
+    """
+    column = get_entity_index(file)
+    return list(re.split(ALL_DELIM, line.strip())[column] for line in file)
+
+def get_entity_index(file) -> int:
+    """
+    Determine which column from a 10X entity file contains valid data
+
+    :param file: (TextIO) open file object
+    :returns: (int)
```
eweitz, line 124:
```suggestion
```

eweitz (approved):
Code looks good overall!

I note two trivial blockers, and many non-blocking maintainability suggestions.

ingest/expression_writer.py
```diff
@@ -0,0 +1,268 @@
+"""expression_writer.py
+extract gene-level expression data from dense/sparse matrix files and process in the
+context of a cluster file's cells.
+this mimics the `expression` data array in visualization API responses
+
+EXAMPLES (must be invoked via ingest_pipeline.py)
+
+dense matrix:
+python3 ingest_pipeline.py --study-id 5d276a50421aa9117c982845 --study-file-id 5dd5ae25421aa910a723a337 \
```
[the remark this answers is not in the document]
bistline: This is definitely baked into ingest pipeline, and I don't feel confident removing this.  My argument against doing so would be that what we're trying to do with image pipeline & DE fall outside of traditional "ingest", and perhaps we shouldn't be invoking those jobs through the ingest harness.  But either way, I can create a ticket to do so.

ingest/expression_writer.py
```diff
@@ -0,0 +1,268 @@
...
+                             render_expression_arrays --matrix-file-path ../tests/data/dense_expression_matrix.txt \
+                             --matrix-file-type dense \
+                             --cluster-file ../tests/data/cluster_example.txt \
+                             --cluster-name 'Dense Example' --render-expression-arrays
```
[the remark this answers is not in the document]
bistline: Same - see above, will ticket.

ingest/expression_writer.py
bistline: Ah thanks - will change.

ingest/writer_functions.py
```diff
@@ -0,0 +1,203 @@
...
+
+def extract_sparse_line(line) -> list:
+    """
+    Process a single line from a sparse matrix and extract values as integers
+
+    :param line: (str) single line from matrix file
+    :returns: (list): values as integers
+    """
+    gene_idx, barcode_idx, raw_exp = line.rstrip().split(' ')
+    return [int(gene_idx), int(barcode_idx), round_exp(raw_exp, 3)]
```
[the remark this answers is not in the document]
bistline: Good catch!

ingest/expression_writer.py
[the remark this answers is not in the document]
bistline: SCP-4737

ingest/expression_writer.py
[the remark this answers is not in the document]
bistline: SCP-4738

ingest/writer_functions.py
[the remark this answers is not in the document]
bistline: [d726dcf](https://github.com/broadinstitute/scp-ingest-pipeline/pull/273/commits/d726dcfeecf909e812fbd0e9cb1c152c1cfbc3b8)

ingest/writer_functions.py
```diff
@@ -0,0 +1,203 @@
...
+    :returns: (int)
+    """
+    first_line = file.readline().strip()
+    file.seek(0) # gotcha to reset pointer to beginning
+    entities = re.split(ALL_DELIM, first_line)
+    return 0 if len(entities) == 1 else 1
+
+def process_sparse_fragment(fragment_name, barcodes, cluster_cells, data_dir):
+    """
+    Process a single-gene sparse matrix fragment and write expression array
```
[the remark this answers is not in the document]
bistline: [d726dcf](https://github.com/broadinstitute/scp-ingest-pipeline/pull/273/commits/d726dcfeecf909e812fbd0e9cb1c152c1cfbc3b8)

ehanna4 (approved):
Functional review only - excited to finally run ingest pipeline locally haha through plenty of trial and error and assistance from Jon was able to run through all the test steps and produce the outputs as advertised

PR #280, opened 2022-11-28
Check that "extract cluster" argument is True

ehanna4:
Previously we were only checking that the argument 'extract cluster' was contained in the argument object to decide whether or not to do the cluster extraction from the AnnData file. However, in playing around locally I realized that that argument is always included in the arguments obj just set  to False or True so now doing that check to only continue with the full ingest when appropriate.

Screen Shot 2022-11-28 at 3 23 09 PM
Screen Shot 2022-11-28 at 3 29 42 PM

### DIFF
--- a/ingest/ingest_pipeline.py
+++ b/ingest/ingest_pipeline.py
@@ -613,7 +613,7 @@ def exit_pipeline(ingest, status, status_cell_metadata, arguments):
                     file_path, study_file_id, files_to_match
                 )
         # for successful anndata jobs, need to delocalize intermediate ingest files
-        elif "extract_cluster" in arguments and all(i < 1 for i in status):
+        elif "extract_cluster" in arguments and arguments.get("extract_cluster") == True and all(i < 1 for i in status):
             file_path, study_file_id = get_delocalization_info(arguments)
             # append status?
             if IngestFiles.is_remote_file(file_path):


eweitz (approved):
Code looks good!

bistline (approved):
Took me a bit to figure out how to validate this but works as advertised.

ingest/anndata_.py
```diff
@@ -18,7 +18,7 @@ def __init__(self, file_path, study_file_id, study_id, **kwargs):
         )
         # If performing cluster extraction, set obsm_keys
         extract_cluster = kwargs.get("extract_cluster")
-        if extract_cluster:
+        if extract_cluster == True:
```
jlchang, line 21: If future us decides to manipulate `obsm_keys` at  `AnnDataIngestor` instantiation, we'll want to check for `True` here too.

ingest/ingest_pipeline.py
```diff
@@ -487,7 +487,9 @@ def ingest_anndata(self):
         )
         if self.anndata.validate():
             self.report_validation("success")
-            if self.kwargs["extract_cluster"]:
+            if self.kwargs["extract_cluster"] == True:
```
jlchang, line 490: Previously, you had noticed that the job errors if `--obsm-keys` isn't set in the command line. I had meant to set up a happy path default so I've provided that here. If FOM requirements mean there will always be at least one expected clustering, we'd set that default here.

jlchang (approved):
Thanks for finding my logic error in the tests for `extract_cluster`. Please scrutinize my additions (I'm sure there was a more appropriate way for me to make my suggestions than a commit - my additions weren't near your change so I wasn't sure how best to share the info, please let me know how I should do so in the future!) 

One of the "suggestions" was really a bugfix I hadn't gotten around to doing and I didn't want to hold up your PR but I also didn't want to forget to make those corrections!

Thanks!

Jean

ingest/ingest_pipeline.py
```diff
@@ -523,7 +525,7 @@ def render_expression_arrays(self):
                 matrix_file_path=self.matrix_file_path,
                 matrix_file_type=self.matrix_file_type,
                 cluster_file_path=self.cluster_file,
-                **self.kwargs
+                **self.kwargs,
```
ehanna4, line 528: @jlchang is this comma intentional here?

PR #281, opened 2022-11-29
Remove cell_type requirement for cell_type__custom (SCP-4835)

jlchang:
This PR removes the declared dependency of `cell_type__custom` on the presence of a corresponding `cell_type` entry in SCP metadata files. When we originally envisioned `cell_type__custom`, we assumed study owners would always be able to find an acceptable “less granular” ontology term to describe cells that needed a custom label. From [Zendesk #298615](https://broadinstitute.zendesk.com/agent/tickets/298615) we learned that study owners would rather not assign a more general `cell_type` annotation for their custom cells and would rather omit a `cell_type` designation for those cells.

An additional automated check in `test_validate_metadata` under `test_custom` automatically tests that this more flexible metadata schema change is adhered to.

To test manually:

Prior state - in the development branch (or prod), see that upload of the file [scp-ingest-pipeline/tests/data/annotation/metadata/convention/valid_cell_type__custom_v2.2.1.txt](https://github.com/broadinstitute/scp-ingest-pipeline/blob/jlc_ctcustom_nodep/tests/data/annotation/metadata/convention/valid_cell_type__custom_v2.2.1.txt) fails validation.

Confirm change - check out this branch and either:

1. Run in your local portal instance - set your local instance to use the following docker image:
`gcr.io/broad-singlecellportal-staging/scp-ingest-jlc_ctcustom_nodep:a177238`
(in Admin config, add `Ingest Pipeline Docker Image`, value type `string` with the above value)
OR
2. Run ingest pipeline via command line
a. Activate the scp-ingest-pipeline repo virtualenv
b. From the ingest directory of the scp-ingest-pipeline repo, perform this setup:
`source ../scripts/setup_mongo_dev.sh <path to your Github token file>`

If you get an authentication error from gcloud, run the following:
`gcloud auth application-default login --no-launch-browser `

c. Run this metadata ingest job from the ingest directory of the scp-ingest-pipeline repo:

```
python ingest_pipeline.py --study-id 5d276a50421aa9117c982845 --study-file-id 5dd5ae25421aa910a723a337 ingest_cell_metadata --cell-metadata-file ../tests/data/annotation/metadata/convention/valid_cell_type__custom_v2.2.1.txt --study-accession SCP4835 --ingest-cell-metadata --validate-convention --bq-dataset cell_metadata_test --bq-table alexandria_convention
```

This fulfills SCP-4835.


bistline (approved):
Thanks for the fast turnaround!  Works as advertised.

ingest/validation/metadata_validation.py
```diff
@@ -0,0 +1,161 @@
+"""Validate input metadata TSV file against metadata convention.
```
eweitz, line 1: Did you do this rename with `git mv`?  

File renames might be Git / GitHub's least well-supported common developer action.  Cases like this make file history in Git discontiguous, and much less feasible to review.

I ~~complained about~~ [explored Git renames](https://broadinstitute.slack.com/archives/C1C22V6FN/p1630673632002800) in #softeng a while ago.  I don't recommend we use the tool noted there; just context for the curious.  The [enhanced GitHub support](https://github.blog/changelog/2022-06-06-view-commit-history-across-file-renames-and-moves/) noted there doesn't seem to have applied here.  The effort to preserve history and enable easy code review in this case is probably not worthwhile.  

I'd just try `git mv` for future cases if you didn't here.  It's not certain to work, but seems like a better bet than raw `mv` or file renaming via editor.

schema/alexandria_convention/Metadata_convention_releases.txt
```diff
@@ -0,0 +1,14 @@
+2.2.1
```
eweitz, line 1: This looks useful!  Suggestions:

* Add dates to releases, to aid debugging
* Use Markdown, to enable clickable links in GitHub via e.g. `[Zendesk #298615](https://broadinstitute.zendesk.com/agent/tickets/298615)`
* Make file name lower case like other files.  Underscore-delimited title case file names are rare in SCP repos.

eweitz (approved):
Code looks good overall!   

I suggest a non-blocking change, and comment on rename issues in Git / GitHub.

ingest/validation/metadata_validation.py
jlchang: This wasn't a rename, I ran into an issue with circular imports. `validate_metadata.py` was supposed to be runnable on the command line. Ingest pipeline only calls functions from `validate_metadata` so the fact that `__main__ `had been changed to invoke `CellMetadata` (with no import from `cell_metadata.py`) meant it could no longer be run via command line to generate reference files to update tests.

To avoid the circular import problem, I duplicated `validate_metadata.py` and created `metadata_validation.py` which has only the command line and argument parsing capabilities that used to be in `validate_metadata.py`. This discussion made me realize I had neglected to update the metadata convention update README to use `metadata_validation.py` instead of `validate_metadata.py`... Going to update that now!

schema/alexandria_convention/Metadata_convention_releases.txt
jlchang: I've updated the file `52a8e8e`, great suggestions - its much more readable and useful now.

PR #284, opened 2022-12-08
Make "cell type (custom)" test data more correct (SCP-4835, SCP-4836)

eweitz:
This trims trailing spaces in a test file, which contributed to UI issues in the new ["cell type (custom)" search facet](https://github.com/broadinstitute/single_cell_portal_core/pull/1685).  

A more robust fix has been ticketed in SCP-4852, but not yet prioritized.  Until that's fixed, "cell type (custom)" values with extra spaces cause duplicate filters to appear in that facet, if the filter has canonical values without trailing space.  

This change simply keeps things clean in the meantime.

It refines SCP-4835 and SCP-4836.


jlchang (approved):
Oof! Thanks for finding those spaces. We should definitely be removing any trailing/leading whitespace (Chomp chomp!) 

Thanks also for the ticket to do that!

bistline (approved):

PR #290, opened 2023-01-21
Enable local development via Docker compose (SCP-4883)

eweitz:
This improves our ability to develop Ingest Pipeline in Apple silicon Macs (e.g. M1), Broad's standard CPU for some time.  In particular, it aims to onboard Michelle to fully contribute to Single Cell Portal.  

The new Docker compose setup reuses patterns from `single_cell_portal_core`.  The updated README here has instructions.  There is known related work (per TODO comments), but the current state seems like a decent working start.

We know that we can _test_ on M1 via local Docker, but my understanding is that no one has been able to _develop_ Ingest Pipeline on an M1.  More background discussion on this is [in #scp-implementation in Slack](https://broadinstitute.slack.com/archives/CBEHTH601/p1673967512735419).  Jon, Michelle and I mobbed on this Thursday; Jon's navigation was essential.

To test:
* Run `./scripts/docker-compose-setup.sh -t ~/.github-token`; where `-t` takes as a value your GitHub token for Vault
* Ensure you see output like the following, and that you enter a Bash shell:
```
### SETTING VAULT TOKEN ###
### SETTING UP ENVIRONMENT ###
Success! You are now authenticated. The token information displayed below
is already stored in the token helper. You do NOT need to run "vault login"
again. Future Vault requests will automatically use this token.

<Some lines excerpted>

Status: Image is up to date for gcr.io/broad-singlecellportal-staging/scp-ingest-pipeline-development:661144f
gcr.io/broad-singlecellportal-staging/scp-ingest-pipeline-development:661144f
root@ac32579d655e:/scp-ingest-pipeline/ingest# 
```

This relates to SCP-4883.


scripts/setup-mongo-dev.sh
```diff
@@ -1,15 +1,17 @@
 #! /bin/bash
```
eweitz, line 1: Would it be OK to remove this script for native development in favor of `scripts/ingest-local-setup.sh` for Dockerized development?  The latter basically adapts the former to use a name and approach that's more consistent with those in Core (i.e. [rails_local_setup.rb](https://github.com/broadinstitute/single_cell_portal_core/blob/217c21ed41e52b4489a5d097cd816091d311745a/rails_local_setup.rb)), and is executed by default in the Docker setup.

If this file can be removed, I would also update references to it in `ingest_pipeline.py` and `README.md`.

README.md
```diff
@@ -131,6 +92,12 @@ pytest --cov=../ingest/
 For more, see <https://docs.pytest.org/en/stable/usage.html>.
 
 ## Testing in Docker
+<!-- 
+Answer in PR 290 and remove this comment:
+Since the "Docker compose" approach described in PR 290 seems more convenient than that 
+described in this "Testing in Docker" section, can I remove this "Testing in Docker" section?
+Or would it be better to scope as part of SCP-4941?
```
eweitz, line 99: This is a question for at least one reviewer.

bistline (approved):
Looks good and work great - this was fun to mob on.  I suggested a few maintainability tweaks and replied to your open questions.

ehanna4 (approved):
Very cool, excited to see this update!

scripts/docker-compose-setup.sh
```diff
@@ -0,0 +1,60 @@
...
+fi
+
+echo "### SETTING UP ENVIRONMENT ###"
+./scripts/ingest-local-setup.sh $VAULT_TOKEN_PATH
+
+docker pull $GCR_IMAGE
+
+# Run the "ingest" service (define in docker-compose-dev.yaml), 
+# then enter the Bash shell
+docker-compose -f docker-compose-dev.yaml run ingest bash
```
jlchang, line 60: Can someone walk me through how `docker-compose-dev.yaml` gets created? I've lost the thread on how this file fits in...

scripts/ingest-local-setup.sh
```diff
@@ -0,0 +1,36 @@
...
+BROAD_USER=`whoami`
+MONGODB_USERNAME='single_cell'
+DATABASE_NAME='single_cell_portal_development'
+MONGODB_PASSWORD=`vault read secret/kdux/scp/development/$BROAD_USER/mongo/user | grep password | awk '{ print $2 }' `
+DATABASE_HOST=`vault read secret/kdux/scp/development/$BROAD_USER/mongo/hostname | grep ip | awk '{ $2=substr($2,2,length($2)-2); print $2 }' ` 
+BYPASS_MONGO_WRITES='yes'
+BARD_HOST_URL="https://terra-bard-dev.appspot.com"
+
+mkdir -p config
+echo "export BROAD_USER=${BROAD_USER}" >| config/.ingest_env
```
jlchang, line 30: Cool - TIL about `>|`, a no-clobber override for file redirection...

scripts/ingest-local-setup.sh
```diff
@@ -0,0 +1,36 @@
...
+echo "export MONGODB_USERNAME=${MONGODB_USERNAME}" >> config/.ingest_env
+echo "export DATABASE_NAME=${DATABASE_NAME}" >> config/.ingest_env
+echo "export MONGODB_PASSWORD=${MONGODB_PASSWORD}" >> config/.ingest_env
+echo "export DATABASE_HOST=${DATABASE_HOST}" >> config/.ingest_env
+echo "export BYPASS_MONGO_WRITES=${BYPASS_MONGO_WRITES}" >> config/.ingest_env
+echo "export BARD_HOST_URL=${BARD_HOST_URL}" >> config/.ingest_env
```
jlchang, line 36: Does `config/.ingest_env` hang around on my filesystem? ie. database password in a plaintext file that is only protected by virtue of it being a dotfile (ie. not listed by default)? or am I overlooking a step in a later script that deletes this file after it is used? Or is it referenced later by other scripts and needs to be kept - if it needs to be kept, should we `chmod 600` the file? (perhaps this caution is overkill for a file on a local developer laptop because `group` and `world` access isn't as relevant as on a shared environment?)

jlchang (changes requested):

scripts/setup-mongo-dev.sh
jlchang: Yes, it looks like `ingest-local-setup.sh` has all the elements of `setup-mongo-dev.sh` so the latter is no longer needed. My only [issue](https://github.com/broadinstitute/scp-ingest-pipeline/pull/290/files#r1085420240) is around any security concerns regarding `config/.ingest_env`

scripts/docker-compose-setup.sh
bistline: The high-level explanation is that `docker-compose-dev.yaml` is a static representation of all the arguments that you would pass to `docker run`.  This allows you to define and bring up/down a service (or collection of services like we do in `single_cell_portal_core`) with a single call to `docker-compose`.  We then pair that with `docker-compose-setup.sh` to automate calling `vault` and pulling the correct Docker image beforehand so that everything is streamlined.

Those two files essentially replicate the following:

```
vault login [credentials_path] >| config/.ingest_env
source config/.ingest_env
docker pull [image name]
docker run [arguments] bash
```

scripts/ingest-local-setup.sh
bistline: That's a great point - it does persist on your machine.  We have the same behavior in `single_cell_portal_core`.  The "problem" is that we can't run any tasks automatically when you `CTRL^C` or type `exit` to kill the service.  We could add that step to a `docker-compose-cleanup.sh` script, but I agree that it's probably "caution overkill" since none of these secrets are for a deployed environment.

scripts/docker-compose-setup.sh
jlchang: Thank you! I _thought_ this was probably where `config/.ingest_env` was being used but didn't want to assume...
and after `docker run` the process is tied up running the instance so you can't use this script to delete `config/.ingest_env`. That's too bad but makes sense.

scripts/ingest-local-setup.sh
jlchang: Okay - as long as we've thought this all through (forgot this is just the creds for my local instance).

jlchang (approved):
Thanks @bistline for walking through the usage of `config/.ingest_env`. I'm satisfied that the security concern I have is limited to non-production environments.

Looking forward to seeing this in action.

scripts/ingest-local-setup.sh
eweitz: Further detail on this good question seems worthwhile.  

In addition to A) reusing that previously-approved pattern Jon notes, we also mitigate security issues with these secrets by B) automatically excluding `config` from Git via [the new `.gitignore` entry](https://github.com/broadinstitute/scp-ingest-pipeline/pull/290/files#diff-bc37d034bad564583790a46f19d807abfe519c5671395fd494d8cce506c42947R3), and C) having our machines configured with `git-secrets` (completely [handled via BITS](https://broadinstitute.slack.com/archives/CADU7L0SZ/p1674575035693219) for recent-ish new machines).

PR #291, opened 2023-01-25
Update sparse and dense matrix examples (SCP-4883)

michelle-vel:
update sparse and dense matrix differential expression tests so that they run with the new file names


bistline (approved):

eweitz (approved):
Code looks good!  

Trivial blocker: please update the PR title and description to replace "tests" with "examples".  Months from now, if I'm researching how our tests have changed and I read this PR, I'll be confused because this code isn't part of the test suite.  These lines kind of similar to tests, but referring to them as examples as we conventionally do (see e.g. line 13 in this file) would be clearer.

PR #292, opened 2023-02-03
Add pairwise DE testing (SCP-4958)

michelle-vel:
Implement A vs B testing for differential expressions following this tutorial: https://nbviewer.org/github/theislab/diffxpy_tutorials/blob/master/diffxpy_tutorials/test/multiple_tests_per_gene.ipynb

Instructions to manually test:
- pip install -r requirements.txt
- python ingest/a_vs_b_de.py

Troubleshooting:
If you receive an error that mentions “Could not find a local HDF5 installation” try the following:
- brew install hdf5
- HDF5_DIR="/opt/homebrew/opt/hdf5" CFLAGS="-I /opt/homebrew/opt/openssl/include" LDFLAGS="-L /opt/homebrew/opt/openssl/lib" GRPC_PYTHON_BUILD_SYSTEM_OPENSSL=1 GRPC_PYTHON_BUILD_SYSTEM_ZLIB=1 pip install -r requirements.txt
- python ingest/a_vs_b_de.py

Current Python version (at the time of writing on my machine) is 3.8. Will add update on other versions if any different instructions are required. 

A Docker shell is not necessary. With the above two commands a pop up should appear of the differential expression graph right away.


ingest/a_vs_b_de.py
```diff
@@ -0,0 +1,65 @@
+#using https://nbviewer.org/github/theislab/diffxpy_tutorials/blob/master/diffxpy_tutorials/test/multiple_tests_per_gene.ipynb
+import anndata
+import matplotlib.pyplot as plt
+import seaborn as sns
```
eweitz, line 4: Blocker: Please add this to `requirements.txt`, so we can install it via `pip install -r requirements.txt`.

In particular, this would be best to add under the ["Dev dependencies" section](https://github.com/broadinstitute/scp-ingest-pipeline/blob/development/requirements.txt#L17), for easier reasoning about its purpose here.  My understanding is Seaborn would only be used for local debugging, e.g. making volcano plots that SCP Ingest Pipeline contributors could use for quick local checks.

ingest/a_vs_b_de.py
```diff
@@ -0,0 +1,65 @@
...
+#arguments:
+#data: Anndata object, data matrix with cells x genes
+
+#grouping: str, column in data, the column that contains cell type labels. alternatively vector containing group labels
+
+#test: type of test, default is z-test. other options: ’wald’, ’lrt’, ’t-test’, ’rank’
+
+#lazy: bool, only possible if test is ztest, if true only evaluated once the user requests pval/coefficients for a specific pair of models
+
+#noise_model: default is nb, specify NONE for wald and t test
```
eweitz, line 35: I don't think it's important to do so now, but we'll presumably want to extract these hard-coded arguments into CLI arguments that can be specified at runtime.  

Let's ensure these to CLI arguments is captured in a downstream ticket.  

After that ticket's created, can you add a TODO comment like the following?

```
# TODO (SCP-<your ticket #>): Extract these to CLI arguments
```

eweitz (changes requested):
Kudos, this is a promising start towards a big enhancement for how SCP does differential expression!  I think it'd be a great topic to discuss at tomorrow's demo.

I note a few blockers, on two aspects of maintainability.  

1.  Others should be able to manually run this code on a terminal by simply running `pip install -r requirements.txt` then `python ...`.

2. The team will need a way to get the data file you're using here.

Comments below go into detail on each of those.  I'm happy to pair, discuss here or via Slack, or help however else.

Finally, a non-blocking suggestion: as we complete this step, it seems like a good time to open a ticket for next steps towards pairwise DE in SCP.

ingest/a_vs_b_de.py
```diff
@@ -0,0 +1,65 @@
...
+adata_file.var_names_make_unique()
+
+#print(adata_file.obs["louvain"])
+
+test = de.test.pairwise(
+    data= adata_file,
+    grouping="louvain",
+    test="t-test",
+    lazy=False,
+    noise_model = None
```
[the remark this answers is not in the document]
jlchang: I'm honestly not sure if my VSCode extensions are relevant (for linting, I believe it's a configuration issue). Here's the output from `code --list-extensions`
DavidAnson.vscode-markdownlint
janisdd.vscode-edit-csv
jhartell.vscode-line-endings
misogi.ruby-rubocop
ms-azuretools.vscode-docker
ms-python.isort
ms-python.python
ms-python.vscode-pylance
ms-toolsai.jupyter
ms-toolsai.jupyter-keymap
ms-toolsai.jupyter-renderers
ms-toolsai.vscode-jupyter-cell-tags
ms-toolsai.vscode-jupyter-slideshow
quarto.quarto
redhat.vscode-yaml
scala-lang.scala
scalameta.metals
shd101wyy.markdown-preview-enhanced
stuart.unique-window-colors
timonwong.shellcheck

Some, like quarto, are for personal projects. Others I can't recall their purpose based on the name of the extension.

IIRC, black is in `requirements.txt` (black 19.3b0) and using black requires a VSCode configuration change

These are the configurations that I could find in my VSCode user configs (not workspace configs) that might be relevant:
"python.linting.enabled": true
"python.formatting.blackPath": "black"
"python.linting.flake8Args": [
  "ignore = E203, E266, E402, E501, W503, W605",
  "max-complexity = 18",
  "select = B,C,E,F,W,T4,B9"
]
"python.linting.ignorePatterns": [
  "**/site-packages/**/*.py",
  ".vscode/*.py"
]
"python.linting.lintOnSave": true

We should probably figure out what/how Black is set and document - would it go in the README? or our onboarding docs? other location(s)?

ingest/a_vs_b_de.py
```diff
@@ -0,0 +1,65 @@
...
+adata_file = sc.read_h5ad(h5ad_file)
+
+adata_file.var_names_make_unique()
+
+#print(adata_file.obs["louvain"])
+
+test = de.test.pairwise(
+    data= adata_file,
+    grouping="louvain",
+    test="t-test",
```
jlchang, line 48: For "A vs all" [we use ScanPy](https://singlecell.zendesk.com/hc/en-us/articles/6059411840027-Exploratory-differential-gene-expression-analysis) with the [Wilcoxon Rank-Sum (Mann-Whitney U Test)](https://en.wikipedia.org/wiki/Mann%E2%80%93Whitney_U_test), a non-parametric test. The "wilcoxon" option for the pairwise comparisons is called ["rank"](https://diffxpy.readthedocs.io/en/latest/api/diffxpy.api.test.pairwise.html). For consistency, we should start with "A vs. B" using "rank" instead of "t-test". 

Blocker: Please adjust this PR to use "rank" and confirm that the syntax here is sufficient for the wilcoxon test. (We saw while pairing that the "wald" test requires a string to specify the generalized linear model it should be testing)

ingest/a_vs_b_de.py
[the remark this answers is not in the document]
eweitz: Thank you Jean!  

> We should probably figure out what/how Black is set and document - would it go in the README? or our onboarding docs? other location(s)?

Agreed, I just created SCP-4973 to track that.

jlchang (changes requested):
The "A vs all" jobs output a file of DE genes, with pvalues. I see that pvalues can be printed to screen. Let's take that a step further and output genes with their pvalues (presumably sorted by significance based on the DE test). The output is necessary to convey the calculated results to the Portal user. For now, let's make that a text file but we should probably have a discussion on how the Portal will want to consume the result so we can adequately spec out a subsequent ticket.

ingest/a_vs_b_de.py
```diff
@@ -0,0 +1,65 @@
...
+
+#grouping: str, column in data, the column that contains cell type labels. alternatively vector containing group labels
+
+#test: type of test, default is z-test. other options: ’wald’, ’lrt’, ’t-test’, ’rank’
+
+#lazy: bool, only possible if test is ztest, if true only evaluated once the user requests pval/coefficients for a specific pair of models
+
+#noise_model: default is nb, specify NONE for wald and t test
+
+h5ad_file = "/Users/mvelyuns/scp-ingest-pipeline/tests/data/anndata/trimmed_compliant_pbmc3K.h5ad"
```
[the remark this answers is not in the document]
eweitz: Thanks!  Michelle noted this yesterday in a Slack chat, and updated the path accordingly on her local branch.

ingest/a_vs_b_de.py
michelle-vel: changing test from "t-test" to "rank" works now!

ingest/de_integration--cell_type__ontology_label--cholinergic_neuron--study--wilcoxon.tsv
```diff
@@ -0,0 +1,188 @@
+	names	scores	logfoldchanges	pvals	pvals_adj	pct_nz_group	pct_nz_reference
```
jlchang, line 1: @michelle-vel This file (and others under `ingest` and `test`) is an output file that does should not go into the repo. Also, the `dotfiles` - .zshrc should be removed from this PR.

jlchang (changes requested):
Blocker - several testing output files (and dotfiles) snuck into this PR, please remove them from the repo.

eweitz (approved):
Looks good!  

I ran the test instructions, and confirm this works.  Thanks for refining the test instructions.  I suggest a few maintainability refinements, nothing blocking.

For convenient reference, here's my output:

Screen Shot 2023-02-15 at 12 30 53 PM

bistline (approved):
Looks good - works as advertised.  One non-blocking suggestion re: comments formatting.

Looking forward - how would we adapt this code if we were to decide that we wanted to implement pairwise differential expression in our `ingest/de.py` library?  Specifically - how much of this code can be reused to output a similar tsv file with results like we have now?

jlchang (approved):
Thanks for removing those extraneous files. Modulo Eric's comment about extracting DE pairwise analysis parameters into CLI arguments, I approve.

PR #293, opened 2023-02-16
Update ingest to use python3.8 (SCP-4547, SCP-4208)

jlchang:
Moving ingest to python3.8 to enable developing on M1 Macs. Updating numpy/scipy in `requirements.txt` enables a working local python environment and requires Python 3.8 which caused build failures in CI due to Python 3.7.5.

This work migrates off CircleCI's [deprecated](https://discuss.circleci.com/t/legacy-convenience-image-deprecation/41034) Legacy Convenience Image `circleci/python:3.7.5-stretch` to use `cimg/python:3.8`

Note for future Python updates: 
when switching convenience images, [increment the cache key name](https://discuss.circleci.com/t/upgrading-to-the-convenience-image-breaks-everything/42746) in `.circleci/config.yml`

QA design: automated
If this works, CI should build and test successfully.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,7 +28,7 @@ jobs:
       - run:
           name: Install Python dependencies
           command: |
-            `python -m site --user-site` -m venv venv
+            python -m venv venv
             . venv/bin/activate
             pip install --upgrade pip
             pip install -r requirements.txt
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,7 +28,7 @@ jobs:
       - run:
           name: Install Python dependencies
           command: |
-            python3 -m venv venv
+            `python -m site --user-site` -m venv venv
             . venv/bin/activate
             pip install --upgrade pip
             pip install -r requirements.txt
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,7 +10,7 @@ orbs:
 jobs:
   build:
     docker:
-      - image: circleci/python:3.7.5-stretch
+      - image: cimg/python:3.8
     resource_class: large
 
     working_directory: ~/scp-ingest-pipeline
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,9 +21,9 @@ jobs:
       # Download and cache dependencies
       - restore_cache:
           keys:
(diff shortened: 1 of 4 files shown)


jlchang:
@eweitz In January you had mentioned wanting to remove `make_toy_data.py` and associated code [slack thread](https://broadinstitute.slack.com/archives/CBEHTH601/p1674497599395109). For this current work, I'm working with a CircleCI config that is no longer installing tabix, a `ingest/genomes` dependency. Will it be necessary to restore tabix as part of the Python upgrade work?

For our future selves, `requirements.txt` has matplotlib pinned to 3.6.3 due to a [two-day-old! bug](https://github.com/scverse/scanpy/issues/2411) with the new Matplotlib 3.7 release.

eweitz:
> Will it be necessary to restore tabix as part of the Python upgrade work?

It doesn't strike me as needed from a product perspective.  If it's easier to remove than to keep, I'd support removal!

eweitz (commented):
Thanks for this infrastructure update!  In addition to enabling M1 development beyond #290, this will also resolve a long-standing (though of uncertain applicability) security issue (#263).

While the [build failure](https://app.circleci.com/pipelines/github/broadinstitute/scp-ingest-pipeline/1784/workflows/e51b23b0-91b9-4ea7-a100-5d8e5c446ced/jobs/1724?invite=true#step-105-36) is being addressed, I have some questions:

* **Why don't we update to Python >= 3.9?**  My vague understanding is that some dependencies fail.  Do you know which those are?  Are there upstream tickets we could watch, to get notified when we could update?  [Python 3.8 end-of-life is 2024-10-14](https://endoflife.date/python).  So we have only 12-18 months before we need to allocate resources to safely avoid EOL.

* **Does this enable native local development?**  Or is Docker still required?  If we can now develop natively, please rearrange the README's "Install" subsections so "Native" is on top, and "Docker" below it.  If it's possible, developing natively (though still via virtualenv) strikes me as faster and simpler than developing in Docker.

Neither of those are blocking concerns.  I plan to approve once the build passes.

jlchang:
Closing this PR in favor of [upgrading to Python 3.10](https://github.com/broadinstitute/scp-ingest-pipeline/pull/294)

PR #295, opened 2023-03-14
Added notebook to create example metadata image (SCP-4963)

michelle-vel:
This notebook creates an example metadata and saves the generated anndata object to an h5ad file, stored in the google bucket at [this link](https://console.cloud.google.com/storage/browser/broad-singlecellportal-staging-testing-data/anndata;tab=objects?prefix=&forceOnObjectsSortingFiltering=false)
Please refer to the two images attached to see what the metadata example looks like.  

Screen Shot 2023-03-14 at 3 33 05 PM

Screen Shot 2023-03-14 at 3 33 42 PM


jlchang (changes requested):
The uploaded file looks like an un-run notebook... Can we upload one that has the expected results?
Screenshot 2023-03-16 at 10 28 46 AM

eweitz (changes requested):
This looks mostly good!  Thanks for taking it on; I'm glad this key image will be largely reproducible via code.  

Below, I note 2 blocking issues and 1 non-blocking suggestion.

Blockers:

**Please add the image(s) to source control**.  Your screenshot is convenient for review, but having the image file itself in the Git repo will make it easier to find and feasible to track changes in over time.  The [classic metadata image](https://raw.githubusercontent.com/broadinstitute/single_cell_portal/master/site_figures/metadata_file/mc_compliant_metadata_image_full.jpg) is a JPG file, but we should probably use PNG as it compresses better for non-photographic images

**Ensure cell names are each contained in 1 non-breaking line**.  For example, the [current draft AnnData metadata image](https://user-images.githubusercontent.com/96444785/225117647-ec107f20-6c50-4dea-a453-08d377f4a18b.png) shows almost all cell names like:

```
AAACATACAACCAC-
1
```

which degrades glanceability and visual quality.  If feasible, we should ensure all cell names in this prominent in-app documentation appear like:

```
AAACATACAACCAC-1
```

1 non-blocking suggestion:

**Visually annotate the image**.  As Jean's noted elsewhere, it'd be quite helpful to not lose the informational details that were in the classic metadata documentation image.  For convenient reference, here's how that other image looks:



Those visual annotations for "biosample_id/donor_id", "Ontology ID", "Ontology label", "Controlled list" and "Unconventional metadata" all provide helpful operational insight for uploaders, and (unlike e.g. the "TYPE" visual annotation) are still relevant for AnnData-formatted files.

The text coloring would be nice but less worthwhile to replicate here.  The visual annotations are both higher value and could presumably be done relatively quickly in e.g. the Preview app in Mac.

I'd be happy to assist with these refinements.  Ping me if pairing would help!

michelle-vel (commented):
added new notebook with results

michelle-vel (commented):
added image with requested edits

jlchang (changes requested):
Two more, hopefully trivial, changes:

1. This PR seems to have two instances of `metadata_image.ipynb`. One at the top level of the repo and one under scripts. Please retain the appropriate instance and keep under `scripts`, deleting the copy at the top level.
2. Similarly, lets create a new top-level directory called `images` where we can keep the image file. The image used for metadata is kept at https://github.com/broadinstitute/single_cell_portal_core/blob/development/app/assets/images/metadata-convention-explainer.jpg, to use similar naming, perhaps the current image can be called anndata-convention-explainer.png.

jlchang (approved):
Updated figure looks great! And thanks for reorganizing the files.

Looking forward to getting your new image into the portal!

PR #297, opened 2023-03-22
Add synthetic example (SCP-4980)

michelle-vel:
Using the diffxpy test-pairwise function from SCP-4958, I added code to create a synthetic file that shows how differential expression pairwise testing might look like. The final file compares between each cell type and shows pvalues and log2FC. For the time being we have eliminated mean expression values, since they are per gene rather than for pairwise comparison, and so do not align with this file format. 

Files stored at [google bucket](https://console.cloud.google.com/storage/browser/broad-singlecellportal-staging-testing-data/anndata/pairwise_DE?pageState=(%22StorageObjectListTable%22:(%22f%22:%22%255B%255D%22))&prefix=&forceOnObjectsSortingFiltering=false)


eweitz (changes requested):
Good start here!  Some high-level blocking issues:

**Omit self-comparison**: The current `output_a_v_b_de` file begins:
```
                            names                                 cell_type                        compared_cell_type          pvals      log2FC
0                            A1BG                                   B cells                                   B cells   0.000000e+00    0.000000
1                        A1BG-AS1                                   B cells                                   B cells   0.000000e+00    0.000000
```

Comparing "B cells" to "B cells" is a no-op, i.e. it will always show no differential expression with a p-value of 0.  The only worthwhile DE comparisons are among different sets, so let's leave out comparisons of a group to itself.

**Output 1 file per pairwise comparison**: The current file outputs all pairwise group comparisons for the given annotation.  The resulting file size is 69 MB raw, and 5.9 MB when compressed via gzip.  

6.0 MB isn't huge, but the common case will be much larger.  Randomly perusing studies in production, the lowest number of groups I saw in DE-eligible annotations was 10.  "[10 choose 2 without repetitions](https://www.omnicalculator.com/statistics/combinations-without-repetition)" gives 45 comparisons.  Your data seems to have 17 comparisons (4 of which compress to smaller than usual due to 0-fill), so it seems a common case would be closer to 3x larger compressed, i.e. 18 MB gzipped.  (I suspect our P75 size with the current model would be >= 30 MB.)

An 18 MB file takes noticeably long to download and (after decompressing to ~200 MB) parse, and unnecessary assuming the UI will rarely show all pairwise comparisons at the same time.  Splitting the single 18 MB into 45 files of 0.4 MB each would make pairwise DE much faster, assuming the UI will generally show DE data for <= all genes for 1 pairwise comparison at a time.

**Add relevant data to output file name**.  Let's build on the pattern of our [one-vs-all DE file names](https://github.com/broadinstitute/scp-ingest-pipeline/blob/development/ingest/de.py#L401), which is:
```
out_file = f'{cluster_name}--{clean_annotation}--{clean_group}--{annot_scope}--{method}.tsv'
```

Assuming my "Output 1 file per pairwise comparison" makes sense, I'd suggest:
```
out_file = f'{cluster_name}--{annotation}--{group_1}--{group_2}--{annot_scope}--{method}.tsv'
```

That would also allow us to omit the group names in each output file, which would substantially reduce even compressed file size.  In other words, we could omit listing out values for "cell_type" and "compared_cell_type" within the file, as it would be in the file name -- and this would further cut down file size (and thus time-to-load at runtime, as well as storage cost.)

**Output TSV**: SCP's canonical flat file format is tab-separated values, i.e. TSV.  This current file delimits by a variable number of spaces.  Let's use a single tab (i.e., `\t`) as delimiter instead.  This will both decrease file size and (more importantly) enable parsing consistent with other flat files.

ingest/a_vs_b_de.py
```diff
@@ -8,13 +8,12 @@
 import scipy.stats
 import scanpy as sc
 import os
-
-
 import batchglm.api as glm
```
jlchang, line 11: If I recall, batchglm was added to see if using those parts of the tutorial would enable pairwise.summary to work. Since the library is not called in this work, we should remove this import for clarity.

jlchang:
This work is to create a synthetic example of how we might perform **internal** pairwise DE data generation so it makes sense to optimize for storage costs and performance.

I'm curious whether the data savings of reducing file size and reformatting data to an internal spec will be worth the overhead if we may be developing a second set of code to handle study-owner contributed custom DE data? Or do we think we'd parse study-owner contributed custom DE data (which I think we expect to be a single file) to create internal files that conform to this spec?

Also, now that the value of DE has been validated, do we need to review whether DE data should continue to be file-based vs ingested into the database for improved performance. Or does the estimated cost (now that we're only doing cell_type-related annotations) of database storage still far exceed the value?

jlchang (changes requested):
Let's make sure to include the validation work for using diffxpy.

Please add the work showing that pairwise DE with `B_Plasma.h5ad` reflects expected differential expression results from [SCP1052](https://singlecell.broadinstitute.org/single_cell/study/SCP1052)'s [publication](https://www.biorxiv.org/content/10.1101/2021.02.25.430130v1), so we capture the method validation as a reference notebook. If you didn't find published expectations for the `B_Plasma` data, we should find a dataset so that validation can be done.

I agree with Eric's suggestions to try adapting the output to our current internal conventions. It makes sense to see what it would take for this data to "fit" existing infrastructure while we wait to see what David Fischer's example file will tell us about how a study owner would organize this data.

eweitz:
Excellent points, Jean!

> do we need to review whether DE data should continue to be file-based vs ingested into the database for improved performance. Or does the estimated cost (now that we're only doing cell_type-related annotations) of database storage still far exceed the value?

For user-perceived performance in viewing and searching a table of DE data, fetching data from files on a bucket is likely fastest.  Fetching from buckets also removes load from the app server and database server, which can result in lower server VM costs by avoiding the need for more powerful VMs (or letting us downsize to lower-power VMs).

Storage cost savings of the bucket strategy for SCP-computed one-vs-all DE are indeed smaller than before, due to narrowing to cell-type-like annotations as you note.  

However, the storage size for SCP-computed pairwise DE -- even just within the current narrow set -- will likely be substantially larger than the prior broad one-vs-all DE.  That's because one-vs-all data grows linearly with the number of groups (_O(n)_), whereas pairwise DE data grows combinatorially (_O(n! / k!(n-k)!)_; n = number of groups, k = 2, i.e. pairwise combinations).  Combinatorial growth will quickly exceed linear growth, even though _n_ is much smaller than before.  That's where the bucket storage savings will really help.

> I'm curious whether the data savings of reducing file size and reformatting data to an internal spec will be worth the overhead if we may be developing a second set of code to handle study-owner contributed custom DE data? Or do we think we'd parse study-owner contributed custom DE data (which I think we expect to be a single file) to create internal files that conform to this spec?

The latter option seems like lower overall development effort and fastest product delivery timing.  I also I agree user-generated DE files will likely be added as a single file.  

In other words, as your second option implies, I think we should accept user DE files as 1 file, then have Ingest Pipeline transform them to per-pairwise-comparison files and load them to the bucket.  The SCP-transformed files could then be directly downloaded in the DE table view for a given pairwise comparison, while the original user DE file could be downloaded from the the study's "Download" tab like other user-uploaded files.

eweitz (approved):
Nice!  This is close to what we'd load into the DE table UI for pairwise comparisons.

The most significant issue I see here is below.  It doesn't need to be fixed now, especially if we no longer consider this PR to be mainly about producing synthetic data for pairwise DE as it would be used in the UI.  But it's worth resolving before we scale up SCP-computed pairwise DE, and perhaps as part of our work to handle custom DE files uploaded by users.

**Output 1 file per combination, not permutation**: For comparing 2 groups (A and B), we should only write out DE data on A vs. B, not also B vs. A.  Given DE data on A vs. B, it seems we can trivially deduce B vs. A.

For example, here's some of the latest synthetic data:

A vs. B, [umap_coords--ClusterName--B cells--Plasma cells--B_cell_cluster--wilcoxon.tsv](https://raw.githubusercontent.com/broadinstitute/scp-ingest-pipeline/04eecd5a0e0b1ce77233b1c7f37bf7ef388eda3a/ingest/umap_coords--ClusterName--B%20cells--Plasma%20cells--B_cell_cluster--wilcoxon.tsv)
```
	names	pvalues	qvalues	log2FC
0	A1BG	0.486519449228373	0.6131885142044022	0.557429356359759
1	A1BG-AS1	0.17405127457430425	0.29135385265304536	-0.7404191774872633
2	A1CF			0.0
3	A2M	0.17822887527874443	0.29708178900033383	0.5162177640511691
```

B vs. A, [umap_coords--ClusterName--Plasma cells--B cells--B_cell_cluster--wilcoxon.tsv](https://raw.githubusercontent.com/broadinstitute/scp-ingest-pipeline/04eecd5a0e0b1ce77233b1c7f37bf7ef388eda3a/ingest/umap_coords--ClusterName--Plasma%20cells--B%20cells--B_cell_cluster--wilcoxon.tsv)

```
	names	pvalues	qvalues	log2FC
0	A1BG	0.486519449228373	0.6131885142044022	-0.557429356359759
1	A1BG-AS1	0.17405127457430425	0.29135385265304536	0.7404191774872633
2	A1CF			-0.0
3	A2M	0.17822887527874443	0.29708178900033383	-0.5162177640511691
```

This indicates p-value and q-value is the same in A vs. B and B vs. A, and log2(FC) in A vs. B is -1x that in B vs. A.

We could reduce pairwise DE storage costs by 2x, and also decrease compute costs, by leveraging that.

**Use same column names as SCP-computed one-vs-all DE**.  A [pre-existing example file](https://github.com/broadinstitute/scp-ingest-pipeline/blob/74a1d9713ccd022661f22c565aa58fc71375cc03/tests/data/differential_expression/sparse/reference/de_sparse_dup_gene--cell_type__ontology_label--endothelial_cell--study--wilcoxon.tsv#L1) has "logfoldchanges" instead of "log2FC", "pvals" instead of "pvalues", and "pvals_adj" instead of "qvalues".  Having different column names for the same metrics would confuse users and internal SCP maintainers.

Eventually we'll want to use e.g. `DifferentialExpression.sanitize_strings` to sanitize groups in SCP-computed pairwise DE file names.  We should also round numbers to the same precision as one-vs-all DE.  

These issues are all non-blocking, likely best resolved as separate later tasks!

bistline (approved):
Looks good - small nit (non-blocking) in that some annotation names are not URL-encoded:
`umap_coords--ClusterName--Plasma cells (long-lived; higher BLIMP1)--Plasma cells--B_cell_cluster--wilcoxon`

Since this is just adding synthetic files, I don't consider this blocking, but should be addressed before we use this in a deployed instance.

jlchang (changes requested):
Hi Michelle - Your previous all-in-one file will be very instructive for thinking about study-owner submitted DE files (because they're going to submit a single file). I don't think the notebook needs to be changed (although, if you have the older notebook, renaming it and add it to the collection of notebooks would be great). If you could add the file to the bucket and indicate that in the PR description, that would be great.

jlchang (approved):
Thanks for adding back the output_a_vs_b_de file to the shared bucket. I think we'll be using that file for the ingest development work!

PR #301, opened 2023-04-18
Refine mtx error messaging for transposed matrices (SCP-5073)

jlchang:
New study owner Sisi Sarkizova reached out in [Zendesk #308520](https://broadinstitute.zendesk.com/agent/tickets/308520) because she was unable to upload her mtx file successfully. In the team [Slack discussion](https://broadinstitute.slack.com/archives/CBEHTH601/p1681241822185889) we observed that the process could detect matrices that need transposition (but we cannot automagically transpose the data for the study owner because future users of the data would expect the mtx file to have its data transposed). We also noted that Sisi likely didn't see the error email so this change is non-blocking.

This PR adds a refined error message when a transposed matrix is detected and enhances the standard feature/cell count mismatch error message. The messaging is automatically tested in test_mtx with tests for non-transposed matrices updated to use the new standard error message and one test in `test_check_valid` for the new transposed matrix error message.

To manually run the automated tests:
1. cd to the `tests` directory
2. `pytest -k test_mtx` should succeed with no errors

### DIFF
--- a/ingest/expression_files/mtx.py
+++ b/ingest/expression_files/mtx.py
@@ -161,6 +161,10 @@ class MTXIngestor(GeneExpression, IngestFiles):
             msg = (
                 f"Expected {expected_barcodes} cells and {expected_genes} genes. "
                 f"Got {actual_barcodes} cells and {actual_genes} genes."
+                f"Uploaded matrix suggests {actual_barcodes} cells and {actual_genes} genes "
+                f"instead of {expected_barcodes} cells and {expected_genes} genes. "
+                f"Please transpose your sparse matrix and re-upload. "
+                f"Transposition is expected for matrices exported from AnnData objects. "
             )
             GeneExpression.log_for_mixpanel(
                 "error", "format:cap:mtx-dimension-mismatch", msg
--- a/ingest/expression_files/mtx.py
+++ b/ingest/expression_files/mtx.py
@@ -161,10 +161,10 @@ class MTXIngestor(GeneExpression, IngestFiles):
             actual_genes == expected_barcodes
         ):
             msg = (
-                f"Uploaded matrix suggests {actual_barcodes} cells and {actual_genes} genes "
+                f"Uploaded matrix suggests {actual_barcodes} columns and {actual_genes} rows "
                 f"instead of {expected_barcodes} cells and {expected_genes} genes. "
                 f"Please transpose your sparse matrix and re-upload. "
-                f"Transposition is expected for matrices exported from AnnData objects. "
+                f"Matrices exported from AnnData objects must be transposed."
             )
             GeneExpression.log_for_mixpanel(
                 "error", "format:cap:mtx-dimension-mismatch", msg
--- a/ingest/expression_files/mtx.py
+++ b/ingest/expression_files/mtx.py
@@ -157,14 +157,24 @@ class MTXIngestor(GeneExpression, IngestFiles):
 
         if (actual_barcodes == expected_barcodes) and (actual_genes == expected_genes):
             return True
-        else:
+        elif (actual_barcodes == expected_genes) and (
+            actual_genes == expected_barcodes
+        ):
             msg = (
-                f"Expected {expected_barcodes} cells and {expected_genes} genes. "
-                f"Got {actual_barcodes} cells and {actual_genes} genes."
                 f"Uploaded matrix suggests {actual_barcodes} cells and {actual_genes} genes "
                 f"instead of {expected_barcodes} cells and {expected_genes} genes. "
                 f"Please transpose your sparse matrix and re-upload. "
                 f"Transposition is expected for matrices exported from AnnData objects. "
+            )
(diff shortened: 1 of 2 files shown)


ingest/expression_files/mtx.py
```diff
@@ -157,10 +157,24 @@ def check_bundle(barcodes, genes, mtx_dimensions):
...
         if (actual_barcodes == expected_barcodes) and (actual_genes == expected_genes):
             return True
+        elif (actual_barcodes == expected_genes) and (
+            actual_genes == expected_barcodes
+        ):
+            msg = (
+                f"Uploaded matrix suggests {actual_barcodes} cells and {actual_genes} genes "
+                f"instead of {expected_barcodes} cells and {expected_genes} genes. "
+                f"Please transpose your sparse matrix and re-upload. "
+                f"Transposition is expected for matrices exported from AnnData objects. "
```
eweitz, line 167: I think a more direct and active voice would decrease cognitive load, e.g.:

```suggestion
                f"Uploaded matrix has {actual_barcodes} cells and {actual_genes} genes "
                f"instead of {expected_barcodes} cells and {expected_genes} genes. "
                f"Please transpose your sparse matrix and re-upload. "
                f"Matrices exported from AnnData objects must be transposed."
```

Also, a tiny copyedit: there is no trailing period in the last sentence of the original message, so I imagine we should also omit it in this one for consistency.

ingest/expression_files/mtx.py
```diff
@@ -157,10 +157,24 @@ def check_bundle(barcodes, genes, mtx_dimensions):
 
         if (actual_barcodes == expected_barcodes) and (actual_genes == expected_genes):
             return True
+        elif (actual_barcodes == expected_genes) and (
+            actual_genes == expected_barcodes
+        ):
```
eweitz, line 162: Having the full clause on one line would make it easier to compare to the other clauses.

```suggestion
        elif (actual_barcodes == expected_genes) and (actual_genes == expected_barcodes):
```

tests/test_mtx.py
```diff
@@ -53,15 +54,31 @@ def test_check_bundle(self):
...
+        barcodes_as_genes = ["cell0", "cell1", "cell3", "cell4"]
+        with self.assertRaises(ValueError) as cm:
+            MTXIngestor.check_bundle(
+                genes_as_barcodes, barcodes_as_genes, mtx_dimensions
+            )
+        expected_msg = (
+            f"Uploaded matrix suggests {len(genes_as_barcodes)} cells and {len(barcodes_as_genes)} genes "
+            f"instead of {expected_barcodes} cells and {expected_genes} genes. "
+            f"Please transpose your sparse matrix and re-upload. "
+            f"Transposition is expected for matrices exported from AnnData objects. "
```
eweitz, line 81: If the other change seems worthwhile:

```suggestion
            f"Uploaded matrix has {len(genes_as_barcodes)} cells and {len(barcodes_as_genes)} genes "
            f"instead of {expected_barcodes} cells and {expected_genes} genes. "
            f"Please transpose your sparse matrix and re-upload. "
            f"Matrices exported from AnnData objects must be transposed."
```

ingest/expression_files/mtx.py
```diff
@@ -157,10 +157,24 @@ def check_bundle(barcodes, genes, mtx_dimensions):
...
+            GeneExpression.log_for_mixpanel(
+                "error", "format:cap:mtx-dimension-mismatch", msg
+            )
+            raise ValueError(msg)
         else:
             msg = (
                 f"Expected {expected_barcodes} cells and {expected_genes} genes. "
-                f"Got {actual_barcodes} cells and {actual_genes} genes."
+                f"Got {actual_barcodes} cells and {actual_genes} genes. "
+                f"Please check the files of your mtx bundle (barcode, feature, mtx) for errors."
```
eweitz, line 177: Nice actionability refinement!

eweitz (approved):
Updated messages look good!  I suggest a few trivial changes, nothing blocking.

ingest/expression_files/mtx.py
jlchang: Great suggestion. Made a more technically correct modification - the uploaded matrix suggests column and row counts that don't match the counts for the cell and gene files so I'm leaving the word "suggests" and changing "cells" -> "columns" and "genes" -> "rows"
`6ba62e0`
Also, updated to be more faithful to generalized nomenclature "genes" -> "features/genes".
`71b083a`

ingest/expression_files/mtx.py
jlchang: Unfortunately that formatting is performed by `Black` which prides itself on being "uncompromising" (ie. no facility for applying formatting exceptions)

tests/test_mtx.py
jlchang: see comment above regarding changes, `6ba62e0` & `71b083a`.

ehanna4 (approved):
code looks good! Thanks for making this improvement, i think it will be really useful for users

PR #303, opened 2023-04-24
Extract processed expression matrix data from Anndata (SCP-4911)

jlchang:
This is a "happy path" implementation of extracting processed gene expression data from h5ad. This extraction currently has no processing of any raw counts data that might be in the `.raw` slot.

Addition of the 'processed_expression' option to the `--extract` flag options array triggers the extractions and loading of processed gene expression data to Mongodb.

To test, run anndata ingest for processed expression matrix:

1. Activate the scp-ingest-pipeline repo virtualenv
2. From the ingest directory of the scp-ingest-pipeline repo, perform this setup:
`source ../scripts/setup-mongo-dev.sh <path to your Github token file>`
3. Run the ingest for a "processed expression matrix"-only extraction:
python ingest_pipeline.py  --study-id 5d276a50421aa9117c982845 --study-file-id 5dd5ae25421aa910a723a337 ingest_anndata --ingest-anndata --anndata-file ../tests/data/anndata/anndata_test.h5ad --extract "['processed_expression']"
4. Check for addition of data arrays to dev MongoDB instance (ie. run `rails c` and perform the following query:
`arrays = DataArray.where(study_id: "5d276a50421aa9117c982845", study_file_id: "5dd5ae25421aa910a723a337")`
5. The number of arrays should be 9 (one all-cells data array, four gene models and four cell expression arrays, one for each gene):
```
irb(main):003:0> arrays = DataArray.where(study_id: "5d276a50421aa9117c982845", 
study_file_id: "5dd5ae25421aa910a723a337")
=> 
#<Mongoid::Criteria                                     
...                                                     
irb(main):004:0> arrays.count
=> 9
```


ingest/anndata_.py
```diff
@@ -127,3 +133,92 @@ def delocalize_extracted_files(file_path, study_file_id, files_to_delocalize):
...
+        # TODO: if raw counts is indicated check that .raw slot is populated
+
+        # Iterate over feature names (for happy path)
+        for feature in self.adata.var_names.tolist():
+            print(f"processing feature: {feature}")
+            feature_expression_series = sc.get.obs_df(self.adata, keys=feature)
+            if feature_expression_series.hasnans:
+                msg = (
+                    f"Expected numeric expression score - "
+                    f"expression data for \'{feature}\' has NaN values"
```
eweitz, line 189: Like elsewhere, let's put the variable part of the message last.  This more robustly handles long values, and generally improves greppability.  Also per convention, let's not demarcate specific values with single quotes in user-facing text.

```suggestion
                    f'expression data has NaN values for feature "{feature}"'
```

tests/test_anndata.py
```diff
@@ -129,9 +144,7 @@ def test_generate_metadata_file(self):
                 "library_preparation_protocol__ontology_label\n",
             ]
             self.assertEqual(
-                expected_line,
-                line,
-                'did not get expected headers from metadata body',
+                expected_line, line, 'did not get expected headers from metadata body'
```
eweitz, line 147: Does Black format this file?  IIUC, [Black would normally transform this to double quotes](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#strings).  As is, I can't discern why this string is single-quoted and strings above are a mix of single- and double-quoted, which is a bit distracting.

ingest/ingest_pipeline.py
```diff
@@ -487,17 +502,27 @@ def extract_from_anndata(self):
         )
         if self.anndata.validate():
             self.report_validation("success")
+            # process matrix data
+            ### TODO: how to associate "raw_count" cells to anndata file
```
eweitz, line 506: Per convention, let's note the SCP Jira ticket this TODO corresponds to, e.g. `TODO (SCP-9876): how...`.

ingest/ingest_pipeline.py
```diff
@@ -487,17 +502,27 @@ def extract_from_anndata(self):
...
+            if self.kwargs.get("extract") and "processed_expression" in self.kwargs.get(
+                "extract"
+            ):
+                self.anndata.process_matrix()
+            # Get cluster extraction parameters and perform extraction
             if self.kwargs.get("extract") and "cluster" in self.kwargs.get("extract"):
                 if not self.kwargs["obsm_keys"]:
-                    self.kwargs["obsm_keys"] = ['X_tsne']
+                    self.kwargs["obsm_keys"] = ["X_tsne"]
+                # TODO: perform check for successful extraction or report failure and exit
```
eweitz, line 515: Please note Jira ticket for TODO, e.g. `TODO (SCP-9876): perform...`".

ingest/ingest_pipeline.py
```diff
@@ -487,17 +502,27 @@ def extract_from_anndata(self):
...
                 for key in self.kwargs["obsm_keys"]:
                     AnnDataIngestor.generate_cluster_header(self.anndata.adata, key)
                     AnnDataIngestor.generate_cluster_type_declaration(
                         self.anndata.adata, key
                     )
                     AnnDataIngestor.generate_cluster_body(self.anndata.adata, key)
+            # Get metadata extraction parameters and perform extraction
             if self.kwargs.get("extract") and "metadata" in self.kwargs.get("extract"):
                 metadata_filename = f"h5ad_frag.metadata.tsv"
+                # TODO: perform check for successful extraction or report failure and exit
```
eweitz, line 525: Please note Jira ticket per convention.

ingest/anndata_.py
```diff
@@ -127,3 +133,92 @@ def delocalize_extracted_files(file_path, study_file_id, files_to_delocalize):
...
+                )
+                GeneExpression.log_for_mixpanel(
```
eweitz, line 191: Blocker: this approach seems like it would log to Mixpanel for _each_ NaN feature.  If there are many such features in a given file, that approach could have a few problems, e.g.:

1. Cause ingest to fail with an unhandled error, due to exceptionally many HTTP requests
2. Make the [equivalent CSFV](https://github.com/broadinstitute/single_cell_portal_core/blob/e54fec3c06c68ac49f799b384341d776f6b59311/app/javascript/lib/validation/expression-matrices-validation.js#L358) impossible to compare to this.  CSFV logs one validation error for _all_ NaN violations in a given file; this seems to log one for _each_ such violation.
3. Yield a very noisy set of error messages for the user
4. Flood Mixpanel with noise, substantially depleting quota

So I think we need to rearrange this error logging to report the error once, at the end, like the CSFV linked above.  That seems not especially trivial, but worth doing here, or at least worth doing before the MVP / public beta for AnnData support.

ingest/anndata_.py
```diff
@@ -127,3 +133,92 @@ def delocalize_extracted_files(file_path, study_file_id, files_to_delocalize):
...
+                feature_expression_series.values > 0
+            ]
+
+            exp_cells = filtered_expression_series.index.tolist()
+
+            untrimmed_exp_scores = filtered_expression_series.values.tolist()
+
+            # trim expression data to three significant digits
+            exp_scores = [round(float(value), 3) for value in untrimmed_exp_scores]
+            # for None value below, consider replacing with gene ID (string)
```
eweitz, line 207: Can we refer to this thing as "feature ID" rather than "gene ID"?  AIUI, these identifiers beyond name in matrices are also potentially useful in SCP for non-gene features, e.g. transcripts.

eweitz (changes requested):
Kudos on getting this big piece of AnnData support working!

I note one blocker, regarding the effects of unbatched logging.  Beyond that, I suggest several refinements for more actionable messages for users, and maintainability.

ingest/anndata_.py
jlchang: Created SCP-5105 for future work to add feature ID checking. ToDo comment in `492d032`

ingest/anndata_.py
jlchang: The first NaN found will cause an unhandled exception so this approach will only log one NaN error per upload.

ingest/ingest_pipeline.py
jlchang: Done `492d032`

ingest/ingest_pipeline.py
jlchang: Done `492d032`

ingest/ingest_pipeline.py
jlchang: Done `492d032`

tests/test_anndata.py
jlchang: Long ago, we wanted to use single quotes but Black uses double. We decided to use `--skip-string-normalization` so Black would ignore quote formatting. In retrospect, not a great idea.

I'm having a bit of bother with Black right now. Once that's addressed, I'd like to do a comprehesive switch to Black's double quote convention so that we have less mixed quote chaos. I'd rather not try to convert quotes in this PR for readability considerations.

ingest/anndata_.py
jlchang: Done `492d032`

ingest/anndata_.py
```diff
@@ -127,3 +133,92 @@ def delocalize_extracted_files(file_path, study_file_id, files_to_delocalize):
...
+
+    @staticmethod
+    def feature_names_unique(features):
+        """Return True if feature names are unique, else false"""
+        if len(features) == len(features.unique()):
+            return True
+        else:
+            # placeholder for function to provide duplicate feature names
+            # non-happy path - option to use gene IDs?
+            msg = f"Duplicate feature name found"
```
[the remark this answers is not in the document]
jlchang: Done `492d032`

ingest/anndata_.py
eweitz: Ah, thanks for helping me see that -- I no longer consider this blocking.  Perhaps helpful:

```suggestion
                # Logged only once per upload, as first NaN causes unhandled exception
                GeneExpression.log_for_mixpanel(
```

eweitz (approved):

ingest/anndata_.py
```diff
@@ -127,3 +135,119 @@ def delocalize_extracted_files(file_path, study_file_id, files_to_delocalize):
...
+                    "error", "content:type:not-numeric", msg
+                )
+                raise ValueError(msg)
+            # capture sparse (only non zero values and their cell IDs)
+            # check mtx.py for all zero gene handling
+            filtered_expression_series = feature_expression_series[
+                feature_expression_series.values > 0
+            ]
+
+            exp_cells = filtered_expression_series.index.tolist()
```
bistline, line 231: How do we think this will perform for large (> 10GB) expression matrices?  Especially if they're particularly dense?  Do you think we run the risk of running out of RAM?

bistline (approved):
Looks good, and works as advertised.  Small question about performance/memory usage under real-world scenarios, but nothing worthy of blocking this from being merged.

PR #307, opened 2023-05-16
Verify and parse input files (SCP-5087)

michelle-vel:
Using the example file we obtained from David Fischer as what a user would input to custom DE, this code parsed the initial file, verified that all the input column headers were correct, and will output an error message if they are not. Then, it further parsed the file to store the values and created individual files in TSV format (as per my previous ticket, one comparison per file). Example files are stored at this [google bucket](https://console.cloud.google.com/storage/browser/broad-singlecellportal-staging-testing-data/anndata/example_custom_DE?pageState=(%22StorageObjectListTable%22:(%22f%22:%22%255B%255D%22))&prefix=&forceOnObjectsSortingFiltering=false).


ingest/validate_de.py
```diff
@@ -0,0 +1,141 @@
...
+names_dict = {}
+all_group = []
+for i in range(len(types)):
+    curr_type = types[i]
+    type_group = []
+
+    for j in range(len(clean_val)):
+        curr_val = clean_val[j][0]
+        comp_val = clean_val[j][1]
+        file_naming = "'" + curr_val + "'" + "_" + "'" + comp_val + "'"
```
bistline, line 80: This will leave `'` characters in the filenames, which is not ideal.  We can safely omit them which will give us more readable filenames.  Also, `f` strings make this much simpler:
```suggestion
        file_naming = f"{curr_val}_{comp_val}"
```

bistline (approved):
Works fine, modulo the issue with the file path at the beginning.  

That being said, while this works as a flat Python script, it would be nice to see something like this broken up into individual methods that perform single functions, and then add a `__main__` section at the bottom that calls the methods and operates on outputs.  The `scripts/convention_to_bq_schema.py` script is a great example of this.  It makes the code more readable as a whole, and also more resuable.

ingest/validate_de.py
eweitz, line 1: For maintainability, I think it'd help to refine this file name to something like `user_de.py` or some such.  That'd help differentiate it from SCP-computed DE, which is programmed in `de.py`.

The fact that we're not only validating user-computed DE but also transforming their one file into many strikes me as another reason to omit `validate` from the filename.

ingest/validate_de.py
```diff
@@ -0,0 +1,141 @@
+import pandas as pd
+import itertools
+import numpy as np
+file_path = "/Users/mvelyuns/scp-ingest-pipeline/ingest/pairwise.csv"
```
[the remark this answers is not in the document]
eweitz: Jon's suggestion seems worthwhile, and broaches an important higher-level topic:

How will this file cohere with the rest of Ingest Pipeline?  The interface between `de.py` and `ingest_pipeline.py` seems worth considering, and presumably replicating here where applicable.  That would help organizational consistency, which helps maintainability.

I don't consider that a blocker for this PR, but I think we should resolve it this sprint or next.

ingest/validate_de.py
```diff
@@ -0,0 +1,141 @@
...
+data = pd.read_csv(file_path)
```
eweitz, line 5: Does this handle TSVs, too?  ([We'll need to](https://docs.google.com/document/d/1R1wBgps4Vx0B98BERGfKIO9H6Qy3MNtSqKTwvAykL3A/edit#).)

ingest/pairwise.csv
eweitz, line 1: Blocker: this file seems to be David's original data.  That's in "[wide format](https://www.statology.org/long-vs-wide-data/)".  Per our latest [plan](https://docs.google.com/document/d/1R1wBgps4Vx0B98BERGfKIO9H6Qy3MNtSqKTwvAykL3A/edit) and the [beginning of our recorded meeting](https://drive.google.com/file/d/19L0-q0zvWeq-vv9tSzL4j9HnUtTCeNQc/view) between Jean, David, you, and me, my understanding is that we'll accept user-computed DE data in "[long format](https://www.statology.org/long-vs-wide-data/)".

I personally don't feel strongly that we should _not also_ support wide format, but I think there's consensus that we would at least support long format.  The rationale is that even though it's not ideal for SCP's visualization engineering purpose, long format is more intuitive and easier for users to supply, and that latter dimension seems more important.

Concretely, I think this means we need to transform this `pairwise.csv` into long format, then split that into the comparison-specific output TSVs as you've done here.  (The output TSVs look basically fine to me; it's just the example input that I think needs to be re-arranged.)

eweitz (changes requested):
This is a substantial feature -- good start here!

I note a few blocking concerns below, and several non-blocking suggestions.  To summarize, the blockers I see are:

1. `pairwise.csv` is in wide format, but we've agreed to support long format
2. Validation seems to require user to compute all comparisons, but we don't want that
3. Error message needs to be more specific and actionable

Issue 3 seems easy (and likely moot), whereas issues 1 and 2 strike me as non-trivial.

I'm free to meet to discuss and pair, if that'd help!

ingest/validate_de.py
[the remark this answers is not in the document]
michelle-vel: both good suggestions! I can certainly break this up into individual functions with a main method and also take a look at the interface between de.py and ingest_pipeline.py.

ingest/type_0_type_1.tsv
eweitz, line 1: For context related to your ongoing work for upload:

[Rails expects](https://github.com/broadinstitute/single_cell_portal_core/blob/f7a1eba618fcfd35a114dd7b83be410dc679b3d4/app/models/differential_expression_result.rb#L82) pairwise DE file names to be formatted like so:

```
      first_label, second_label = Naturally.sort([label, comparison]) # comparisons must be sorted alphabetically
      values = [cluster_name, annotation_name, first_label, second_label, annotation_scope, computational_method]
      ...
     basename = values.map { |val| val.gsub(/\+/, 'pos').gsub(/\W/, '_') }.join('--')
     "#{basename}.tsv"
```

I believe pre-existing Ingest Pipeline code has some handling for that, but it'll need to be imported / accounted for in `user_de.py` or wherever your code to upload ends up residing.  

In other words, instead of having a file name like:

`type_0_type_1.tsv`

you'd need to output a file name like:

`cluster_umap_txt--General_Celltype--type_0--type_1--wilcoxon.tsv`

where `cluster_umap_txt` would be the (santized, hypothetical) value for a `cluster_name` parameter, `General_Celltype` for `annotation_name`, `type_0` for `group_a`, `type_1` for `group_b`, and `wilcoxon` for `computational_method`.

This doesn't need to be done now, but hopefully helps as context for your other ongoing work.  Does all that sound right, @bistline?

eweitz (approved):
Code looks much better, kudos!

I note one trivial blocker -- see comment beginning "Blocker:".  I also suggest a handful of non-blocking maintainability refinements, and surface some cross-systems context relevant for the downstream Rails client.

ingest/user_de.py
```diff
@@ -0,0 +1,178 @@
...
+import numpy as np
+import sys
+
+#input file name as CLI
+#python ingest/validate_de.py /Users/mvelyuns/scp-ingest-pipeline/ingest/pairwise.csv
+
+if len(sys.argv) > 1:
+    file_path = sys.argv[1]
+else:
+    file_path = "/Users/mvelyuns/scp-ingest-pipeline/ingest/pairwise.csv"
```
bistline, line 11: This really should just be a relative path of `ingest/pairwise.csv` but accepting the argument above fixes this.

ingest/user_de.py
```diff
@@ -0,0 +1,178 @@
...
+    rest = data[data.columns[1:]]
+    col = list(rest.columns)
+    col_values = {}
+    for i in col:
+        col_values[i] = rest[i].tolist()
+    return col, genes, rest
+
+#note: my initial files had pval, qval, log2fc. David's files have qval, mean, log2fc. For the purposes of this validation I will be using his column values/formatting. If we choose to remove mean again this code can be changed
+
+#question: should user also input type names to not rely solely on parsing?
```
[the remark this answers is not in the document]
michelle-vel: should this code output an error message if the metrics are not qval, log2fc, and mean at this point? doing so will ensure we only receive those metrics that we support at the moment, but not having that will make it very easy to add additional metrics in the future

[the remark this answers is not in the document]
eweitz: Let's validate that at least log2fc and qval exist as column headers.  

The [DE explore UI](https://github.com/broadinstitute/single_cell_portal_core/pull/1768#user-content-video) currently supports showing 3 columns: gene name, and two numerical metrics (log<sub>2</sub>(FC) and Adj. p-value).  Supporting a larger set of columns (even arbitrarily named ones) wouldn't be a huge effort, but isn't trivial either, so ensuring at least those 2 metrics for log2fc and qval exist seems adequate as a way to ensure the UI doesn't break while also enabling users to include more data in their original file if they like.

(The fact that "log2fc" is superficially different from existing DE's "log2foldchanges" and "qval" differs from "pvals_adj" is fine and expected, FWIW.)

PR #309, opened 2023-06-06
Add manifest file and support for one vs rest (SCP-5087)

michelle-vel:
Making some adjustments to the last [PR](https://github.com/broadinstitute/scp-ingest-pipeline/pull/307) I submitted to add support for one vs rest file comparisons and generating an additional manifest file with the output. 

This PR includes both possible variations of the manifest file. In practice they will both be called manifest.tsv but so I could upload both I changed the name of one of them to manifest_one_vs_rest.tsv. This PR also includes all the individual files generated by one vs rest comparison, as well as a sample one_vs_rest.csv file. This file is not based in real data, I simply edited the pairwise.csv file we received from David Fischer to make it into one vs rest format. 

This PR also includes a change which raises an exception if the columns don't have at least log2fc and qval as comparison values.


ingest/user_de.py
```diff
@@ -170,15 +177,31 @@ def generate_individual_files(col, genes, rest, groups, clean_val, qual):
 
 # final result: individual files for each comparison
 
+def generate_manifest(clean_val, qual):
```
bistline, line 180: Blocker: I believe what we had discussed was a tab-separated file where each "group" observation was a separate _line_, not a _column_.  This will make processing it much simpler back in the portal.  So the file should read:
```
type_0
type_1
...
```

Also, if we use that format, we don't need two manifests.  We can have a single two-column manifest where a line with one value is considered one vs. rest, and a line with two values is pairwise.  E.g:
```
type_0
type_1
type_2
type_0  type_1
type_0  type_2
type_1  type_2
```

ingest/user_de.py
```diff
@@ -170,15 +177,31 @@ def generate_individual_files(col, genes, rest, groups, clean_val, qual):
...
     long_format = convert_wide_to_long(data)
     wide_format = convert_long_to_wide(long_format)
-
     data_by_col = get_data_by_col(data)
     col, genes, rest = data_by_col
-
     groups_and_props = get_groups_and_properties(col)
     groups, clean_val, qual = groups_and_props
+    generate_manifest(clean_val, qual)
+    generate_individual_files(col, genes, rest, groups, clean_val, qual)
```
bistline, line 205: Was using the correct format for for the individual filenames supposed to be a part of this PR, or was that for subsequent work to push the file back to the bucket?  Whenever we decide that work needs to be done, the filenames need to follow [this structure](https://github.com/broadinstitute/scp-ingest-pipeline/blob/development/ingest/de.py#L393).
```
out_file = f'{cluster_name}--{clean_annotation}--{clean_group}--{annot_scope}--{method}.tsv'
```

bistline (changes requested):
Nice work - see blocker comment about the format of the manifest.  We want values in rows, not columns.

ingest/one_vs_rest.csv
```diff
@@ -0,0 +1,41 @@
+,'type_0'_'rest'_qval,'type_0'_'rest'_log2fc,'type_0'_'rest'_mean,'type_1'_'rest'_qval,'type_1'_'rest'_log2fc,'type_1'_'rest'_mean,'type_2'_'rest'_qval,'type_2'_'rest'_log2fc,'type_2'_'rest'_mean,'type_3'_'rest'_qval,'type_3'_'rest'_log2fc,'type_3'_'rest'_mean
```
eweitz, line 1: Can we omit the single quotation marks in this row, and the other example data for user DE?  

It's reflective of the user's original data, but I suspect it represent be the majority of the data we'll ultimately get from users.  Having an automated test that we can handle group names enclosed in quotes would help ensure we keep supporting this likely less common case.

ingest/user_de.py
```diff
@@ -1,15 +1,18 @@
...
 # input file name as CLI
 # example:
-# python ingest/validate_de.py /Users/mvelyuns/scp-ingest-pipeline/ingest/pairwise.csv
+# python ingest/user_de.py /Users/mvelyuns/scp-ingest-pipeline/ingest/pairwise.csv
 
 if len(sys.argv) > 1:
     file_path = sys.argv[1]
 else:
     file_path = "ingest/pairwise.csv"   
+    #file_path = "ingest/one_vs_rest.csv"
```
eweitz, line 14: Before SCP-5086 and SCP-5087 are marked done, let's refactor `user_de.py` to remove this block and be called more like `de.py` and the other modules in Ingest Pipeline.  In other words, we'll need to ensure your `user_de.py` code can be called via a terminal command to `ingest_pipeline.py` that includes something like those CLI arguments we discussed late last week (i.e., something like the [example here](https://github.com/broadinstitute/scp-ingest-pipeline/blob/development/ingest/ingest_pipeline.py#L56), though the particular arguments will differ).

I don't think that is essential for this PR.  But I do think we'll need it before these two tickets are complete, because a main intent of them is to unblock Rails.

ingest/user_de.py
```diff
@@ -170,15 +177,31 @@ def generate_individual_files(col, genes, rest, groups, clean_val, qual):
...
 
+def generate_manifest(clean_val, qual):
+    """
+    create manifest file of each comparison in the initial data
+    if the comparison is with rest, rest is omitted and just the type is written
+    """
+    file_names = []
+    for i in clean_val:
+        cleaned_list = [ x for x in i if x != "rest" and x not in qual]
+        file_name = '_'.join(cleaned_list)
```
eweitz, line 188: Blocker: please ensure each pairwise comparison line has both groups in their natural sort order.  

For example, if the groups in the comparison are "type_2" and "type_10", the manifest line should list them as 

* type_2     type_10

and not

* type_10     type_2

we'll want to ensure that's so even if the user lists "type_10" before "type_2" in their DE file.  Having the test file contain "type_10" before "type_2" would be a good way to ensure we manually test that.  This would also be a great motivating example to start on automated tests for user DE in Ingest Pipeline!

eweitz (changes requested):
Another good step towards supporting user-uploaded DE!  

I suggest one non-trivial blocker, one trivial suggestion for refinement, and note a high-level change we'll need before this ticket and its companion ticket are done.

Another high-level note, somewhat related to Jon's: 

The new synthetic data is _only_ one-vs-rest comparisons.  This nicely demonstrates isolated support for that class of DE.  The previous data only pairwise comparisons.  

An additional scenario I think we'll need to support is a user supplying a DE file that has _both_ one-vs-rest _and_ pairwise comparisons.  A [key result](https://docs.google.com/spreadsheets/d/1Gm5-_rEJzJvIkLIIziP3nz80dIB7H2C3S0vyvA4o7E4/edit#gid=0) for this quarter is "end-to-end support for users uploading both 1-vs-all and pairwise comparisons".  So while isolated support for one-vs-rest and pairwise might suffice, I suspect the intent is to support both comparisons in the same annotation, and thus the same user-uploaded DE file.

ingest/manifest_one_vs_rest.tsv
```diff
@@ -0,0 +1 @@
+type_0	type_1	type_2	type_3
```
eweitz, line 1: Blocker: please see [Jon's note](https://github.com/broadinstitute/scp-ingest-pipeline/pull/309#discussion_r1220103112) about not using two manifest files.  Let's use just one.  His example in that link is a good reference of how the single manifest file needs to look, and how to distinguish one-vs-rest and pairwise comparisons.

ingest/user_de.py
```diff
@@ -164,21 +212,70 @@ def generate_individual_files(col, genes, rest, groups, clean_val, qual):
...
-    wide_format = convert_long_to_wide(long_format)
+def generate_manifest(clean_val, clean_val_p, qual):
+    """
+    create manifest file of each comparison in the initial data
+    if the comparison is with rest, rest is omitted and just the type is written
+    """
+    file_names_one_vs_rest = []
+    for i in clean_val:
+        cleaned_list = [ x for x in i if x != "rest" and x not in qual]
+        file_name_one = '_'.join(cleaned_list)
```
eweitz, line 234: Blocker: please see Jon's note in his review about [output filename format](https://github.com/broadinstitute/scp-ingest-pipeline/pull/309#discussion_r1220107692).  

### Delimiter
The delimiter for each "segment" (like "type_0") of the file name must be a string that does not appear within any segment.  

Here, the current delimiter is "_".  Underscores can also appear internal in a filename segment, e.g. "type_3".  So we need to use a different string as a delimiter.

Ingest Pipeline conventionally uses two hyphens for such delimiters.  Let's apply that here.

```
        clean_comparison = '_'.join([ x for x in i if x != "rest" and x not in qual])
```

### Context segments
Above in this file, you have placeholders.  Here they are, with more realistic values:
```
cluster_name = "All_Cells_UMAP"
clean_annotation = "General_Celltype"
annot_scope = "study"
method = "wilcoxon"
```

As Jon notes in that link, the output filenames need to contain these values (here is replace `clean_group` from Jon with a clearer variable name `clean_comparison`).  This would be something like:

file_name_one = f"{cluster_name}--{clean_annotation}--{clean_comparisons}--{annot_scope}--{method}"

More concretely, we need file names like:

* Pairwise: `All_Cells_UMAP--General_Celltype--type_2--type_10--study--wilcoxon.tsv`

* One-vs-rest: `All_Cells_UMAP--General_Celltype--type_1--study--wilcoxon`
  * ^ Note how "rest" is omitted in `clean_comparisons`.

eweitz (changes requested):
Looking better!  I note some issues below.

To help clarify each of the comments in my review here is noticed, can you write a brief reply in each comment when it's addressed?

As always, I'm free to discuss in Slack and pair program if that'd help.

ingest/user_de.py
michelle-vel: file names changed, rest omitted from file name

ingest/manifest_one_vs_rest.tsv
michelle-vel: correct manifest file uploaded - wrong one was provided initially, my bad!

ingest/user_de.py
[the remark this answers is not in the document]
michelle-vel: changed output file name format

ingest/user_de.py
```diff
@@ -18,31 +28,47 @@ def convert_wide_to_long(data):
     long = pd.melt(data, id_vars='genes', var_name = "comparisons", value_vars= ls)
     return long
 
-def convert_long_to_wide(data):
+def get_long(data):
```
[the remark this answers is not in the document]
michelle-vel: changed names

ingest/user_de.py
```diff
@@ -164,21 +212,70 @@ def generate_individual_files(col, genes, rest, groups, clean_val, qual):
         arr = np.array(file_d[i])
         t_arr = arr.transpose()
         inner_df = pd.DataFrame(data = t_arr, columns = qual)
+
+        #inner_df.to_csv(f'{cluster_name}--{clean_annotation}--{clean_group}--{annot_scope}--{method}.tsv'.format(cluster_name, clean_annotation, clean_group, annot_scope,i), sep ='\t')
         inner_df.to_csv(f"ingest/{i}.tsv".format(i), sep ='\t')
+
         final_files_to_find.append("ingest/{}.tsv".format(i))
```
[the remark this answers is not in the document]
michelle-vel: changed file naming

ingest/user_de.py
```diff
@@ -96,6 +127,21 @@ def check_group(names_dict, name):
         if name in names_dict[i]:
             return i
 
+def verify_sorting(ls):
+    """
+    helper function to sort either alphabetically or by the numbers in the list
+    this should take in only a list of two, ex (type1 type2)
```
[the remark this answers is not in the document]
michelle-vel: added to docstring

PR #310, opened 2023-06-26
Finish integrating author DE into Ingest Pipeline 

michelle-vel:
Extend implementation for author DE into the Ingest Pipeline codebase, by adding it to ingest_pipeline.py and adding a new parser to cli_parser.py. The structure of doing so is very similar to de.py and how automated DE files are generated and pushed to google bucket from the ingest pipeline. The end result is individual comparison files being written, for pairwise and rest DE, and uploaded to google bucket. See an example bucket [here](https://console.cloud.google.com/storage/browser/fc-3dab559a-a5ce-42a6-96e7-1e04228c10b8/_scp_internal/differential_expression;tab=objects?pageState=(%22StorageObjectListTable%22:(%22f%22:%22%255B%255D%22,%22s%22:%5B(%22i%22:%22objectListDisplayFields%2FtimeLastModified%22,%22s%22:%221%22),(%22i%22:%22displayName%22,%22s%22:%220%22)%5D))&prefix=&forceOnObjectsSortingFiltering=true). 

### Test 

1.  Go to your local SCP
2. Find "Human milk - differential expression" study
3. Go to its Google bucket
4. Upload the example author DE file from `scp-ingest-pipeline/tests/data/author_de/author_de_test_data_human_milk_All_Cells_UMAP_General_celltype.csv` to the top-level GCS bucket directory
5. Run the following command, replacing `fc-febd4c65-881d-497f-b101-01a7ec427e6a` with your bucket ID.

```
python ingest_pipeline.py --study-id addedfeed000000000000000 --study-file-id dec0dedfeed1111111111111 ingest_differential_expression --annotation-name General_Celltype --annotation-type group --annotation-scope study --cluster-name cluster_umap_txt --study-accession SCPdev --ingest-differential-expression --differential-expression-file gs://fc-febd4c65-881d-497f-b101-01a7ec427e6a/author_de_test_data_human_milk_All_Cells_UMAP_General_celltype.csv --method wilcoxon
```

6.  Back in your GCS bucket, confirm you have new files in `_scp_internal/differential_expression/`.


ingest/cli_parser.py
```diff
@@ -331,6 +331,63 @@ def create_parser():
...
+        "--cluster-file",
+        required=True,
+        help="Path to cluster file.",
+    )
+
+    parser_ingest_differential_expression.add_argument(
+        "--annotation-file",
+        required=True,
+        help="Absolute or relative path to cell metadata or cluster file of annotations.",
+    )
```
eweitz, line 383: This isn't strictly needed, and would be ideal to remove, but doesn't seem worth doing now.

ingest/ingest_pipeline.py
```diff
@@ -719,6 +746,17 @@ def main() -> None:
         arguments["cell_metadata_file"] = arguments["annotation_file"]
         # IngestPipeline initialization expects "name" and not "cluster_name"
         arguments["name"] = arguments["cluster_name"]
+
+
+    if "ingest_differential_expression" in arguments:
+        # DE may use metadata or cluster file for annots BUT
+        # IngestPipeline initialization assumes a "cell_metadata_file"
+        arguments["cell_metadata_file"] = arguments["annotation_file"]
```
eweitz, line 754: This is a vestige from `de.py` handling.  We should theoretically be able to remove `cell_metadata_file` / `annotation_file`, but that doesn't seem worth doing here.  Retaining it for now seems like reasonable tech debt.

eweitz (approved):
Code looks good!  I ran this locally for a local study's bucket, and it worked.

ingest/cli_parser.py
```diff
@@ -331,6 +331,63 @@ def create_parser():
...
+
+    parser_ingest_differential_expression.add_argument(
+        "--ingest-differential-expression",
+        required=True,
+        action="store_true",
+        help="Indicates that author differential expression analysis should be invoked",
+    )
+
+    parser_ingest_differential_expression.add_argument(
```
bistline, line 346: It doesn't break anything, but I don't know that we need `--study-accession` for DE jobs.  Typically it's only needed for metadata jobs since we need that value in BigQuery.

ingest/cli_parser.py
```diff
@@ -331,6 +331,63 @@ def create_parser():
...
+        "--method", default="wilcoxon", help="method for DE"
+    )
+
+    parser_ingest_differential_expression.add_argument(
+        "--cluster-name", required=True, help="study owner-specified cluster name"
+    )
+
+    # TODO: after end to end support, remove this argument
+    parser_ingest_differential_expression.add_argument(
+        "--cluster-file",
```
bistline, line 374: Blocker: we should remove this and `--annotation-file` since they're not used, and currently not included from the Rails side as arguments.  We would be able to figure them out, but given that they're superfluous it seems preferable to remove them.

ingest/cli_parser.py
bistline: Given that they're marked as `required` and also not referenced anywhere, I think we should remove them.  It will require extra effort to source values from the Rails side to set for these, and given that they're not referenced anywhere it does seem better to remove them now.

bistline (changes requested):
Looks good modulo the comment re: `annotation_file` and `cluster_file`.  These are marked as required but not used in `author_de.py`, and will require extra integration effort from Rails to set values for these (they're currently not collected in the upload UI, though it would be possible to determine them).  Given that I think it's preferable to remove them now as it will require PRs in separate repositories to remove later.

ingest/cli_parser.py
eweitz: You're right it's not needed, but having study accession available at a glance in logs, Sentry, etc. substantially improves debuggability.

ingest/author_de.py
```diff
@@ -0,0 +1,405 @@
+"""Ingest differential expression uploaded by authors, i.e. study owner / editor
+
+EXAMPLE:
+python ingest_pipeline.py --study-id addedfeed000000000000000 --study-file-id dec0dedfeed1111111111111 ingest_differential_expression --annotation-name General_Celltype --annotation-type group --annotation-scope study --annotation-file ../tests/data/differential_expression/de_dense_cluster.tsv --cluster-file gs://fc-febd4c65-881d-497f-b101-01a7ec427e6a/cluster_umap.txt --cluster-name cluster_umap_txt --study-accession SCPdev --ingest-differential-expression --differential-expression-file author_de_test_data_human_milk_All_Cells_UMAP_General_celltype.csv
```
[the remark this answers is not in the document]
eweitz: Good catch!  Added in 7563c83.

ingest/cli_parser.py
eweitz: Done!

ingest/cli_parser.py
eweitz: Done.  This was referenced downstream, but with some effort I removed it.

bistline (approved):

ingest/ingest_pipeline.py
eweitz: (This tech debt is now resolved.)

PR #319, opened 2023-09-01
Author DE: Fix pairwise columns, support TSV (SCP-5292)

eweitz:
This fixes a scientific bug in transforming differential expression data, and improves robustness in a few areas.

Previously, columns would become misordered when the author DE file contained multiple pairwise comparisons that started with the same group.  For example, if the author DE file had pairwise comparisons for, say, "B cells and macrophages" _and_ "B cells and T cells", then the processed output for "B cells and macrophages" could contain metrics for "B cells and T cells".  Now, output is reliably ordered, and test data has been updated accordingly.

This also includes some scaffolding code for more flexible headers.  I found the bug while developing that, so -- as you can see in `de_utils.py` -- most of that code is commented out.  The current code retains the "basic" validation to ensure the author DE table explore UI doesn't break.

Finally, TSV data is now accepted as input.

To manually test:

1.  Run `python ingest_pipeline.py --study-id addedfeed000000000000000 --study-file-id dec0dedfeed1111111111111 ingest_differential_expression --annotation-name General_Celltype --annotation-type group --annotation-scope study --cluster-name cluster_umap_txt --study-accession SCPdev --ingest-differential-expression --differential-expression-file gs://fc-febd4c65-881d-497f-b101-01a7ec427e6a/author_de_test/lfc_qval_scanpy-like.csv --method wilcoxon`
2. Run `ls -latr cluster_umap_txt--General_Celltype--B_cells--CSN1S1_macrophages--study--wilcoxon.tsv`
3. Confirm you see that file listed, with a timestamp from a few moments ago
4. Run `head -n 2 cluster_umap_txt--General_Celltype--B_cells--CSN1S1_macrophages--study--wilcoxon.tsv`
5. Confirm you see expected output of author DE pipeline
```
	genes	logfoldchanges	qval	mean
0	ACE2	0.685269917070053	0.446924681159184	0.727719408271226
```
6.  Confirm it matches the corresponding row from the raw, untransformed author DE file: run `grep "ACE2,B cells,CSN1S1 macrophages" /tmp/author_de_test%2flfc_qval_scanpy-like.csv`
7. Confirm you see `ACE2,B cells,CSN1S1 macrophages,0.6852699170700530,0.44692468115918400,0.7277194082712260`.

This satisfies SCP-5292.


ingest/author_de.py
```diff
@@ -220,28 +311,42 @@ def execute(self):
         metrics = []
         file_path = self.author_de_file
 
-        data = pd.read_csv(file_path)
+        # TODO:
+        #   - Throw well-formatted error if file type not in ALLOWED_FILE_TYPES
+        #   - Consider if / how this merges with centralized handling in ingest_files.py
+        file_type = IngestFiles.get_file_type(file_path)[0]
+        if file_type == 'text/tab-separated-values':
```
bistline, line 318: Out of scope: I know we've had issues generally with ingest and sniffing delimiters.  In the Rails codebase, we're able to read & split lines on multiple delimiters by passing a regex like `/[,\t]/`.  Does `read_csv` have similar support for this?

bistline (approved):
Looks good - one non-blocking question about delimiters when using `read_csv`.

ingest/author_de.py
eweitz: > Does read_csv have similar support for this?

It does!  From [`pandas.read_csv` docs](https://pandas.pydata.org/docs/reference/api/pandas.read_csv.html), emphasis mine:


> ***sep str, default ‘,’***
Character or regex pattern to treat as the delimiter. If sep=None, the C engine cannot automatically detect the separator, but the Python parsing engine can, meaning the latter will be used and automatically detect the separator from only the first valid row of the file by Python’s builtin sniffer tool, csv.Sniffer. In addition, **separators longer than 1 character and different from '\s+' will be interpreted as regular expressions** and will also force the use of the Python parsing engine. Note that regex delimiters are prone to ignoring quoted data. Regex example: '\r\t'.
>
> ***delimiter str, optional***
Alias for sep.

I'll give that a try in my next PR.

ingest/de_utils.py
```diff
@@ -0,0 +1,85 @@
+"""Utilities for differential expression
+"""
+
+import re
+
+
+# def get_size(metric):
```
ehanna4, line 7: How come these functions are commented out rather than deleted? Or is the comment below "# TODO: Delete this function once UI can handle flexible metrics" related?

ehanna4 (approved):
Overall code looks good, one question around commenting out vs deleting code.

ingest/de_utils.py
eweitz: Ah, I should've clarified.  The commented-out code here would mostly be uncommented when that TODO is done.

PR #322, opened 2023-09-08
Author DE: support Seurat and other formats, fix numerical groups (SCP-5296)

eweitz:
_If you haven't, please first review #321_.

---

This adds author DE support for output from [Seurat's FindAllMarkers()](https://satijalab.org/seurat/articles/pbmc3k_tutorial.html#finding-differentially-expressed-features-cluster-biomarkers) function, which a user recently uploaded.

### Support Seurat and other formats
More generally, these changes implement flexible author DE headers to support _any_ long-header-format DE file uploaded by the user, extending [early parameterization under review](https://github.com/broadinstitute/scp-ingest-pipeline/pull/321).  This is achieved by removing hard-coded assumptions about the names and order of the input headers "gene", "group", and "comparison_group".  The "comparison_group" header is also now optional, easing upload of more-natively-formatted data from e.g. default Seurat analysis that _only_ contains one-vs-rest comparisons (and no pairwise comparisons).  

So, for example, this helps enable authors to upload not only Seurat FindAllMarkers() output -- which has a non-pairwise structure and headers that differ from our original expectations -- but _any_ DE file that isn't in wide format.  This means that if Seurat or Scanpy updates header strings they output, then author DE ingest won't break.  The headers for "gene", "group", and (if applicable) "comparison_group" will be provided by the author in the upload UI.  Specifying those header names will typically require no author effort, as they'll be inferred upon file selection [like "Size metric" and "Significance metric"](https://github.com/broadinstitute/single_cell_portal_core/pull/1874), instantly reviewable at a glance, and overridable in 2 clicks if needed.

The output files used by the DE UI remain the same as before.  

### Fix numerical groups
Finally, this also fixes a lurking bug with ingesting differential expression comparisons for groups that have numerical labels (but are indeed categorical).  Such groups are often seen in early e.g. Leiden or Louvain clustering analysis.  

For example, previously, author DE ingest had bugs if the annotation had groups like "0", "1", "2", "3", ..., "20".  DE output for groups like "10" and "11" would simply be missing, due to subtle non-robustness in the `check_group` function.  This bug could have also manifested beyond numerical labels, albeit less commonly. 

Now the bug is fixed.  Transformation is robust for author DE involving numerical groups.

### Test
Automated tests have been updated. To manually test:

Run:
> python ingest_pipeline.py --study-id addedfeed000000000000000 --study-file-id dec0dedfeed1111111111111 ingest_differential_expression --annotation-name General_Celltype --annotation-type group --annotation-scope study --cluster-name cluster_umap_txt --study-accession SCPdev --ingest-differential-expression --differential-expression-file ../tests/data/author_de/seurat_findallmarkers_one-vs-rest.csv --method wilcoxon --genes-header gene --group-header cluster --comparison-group-header None --size-metric avg_log2FC --significance-metric p_val_adj; head cluster_umap_txt--General_Celltype--10--study--wilcoxon.tsv

Confirm you see:
```
	genes	avg_log2FC	p_val_adj	pct.2	pct.1	p_val
0	CRP	1.082893128	0.0	0.1364	0.3268	0.0
1	LGALS3	1.399143463	0.0	0.29848	0.4284	0.0
2	PIK3CA	0.7331456728	7.070000000000001e-273	0.34236	0.67932	3.49e-277
```

This satisfies SCP-5296.

### DIFF
--- a/ingest/author_de.py
+++ b/ingest/author_de.py
@@ -35,25 +35,37 @@ def sort_comparison(groups):
         return sorted(groups)
 
 
-def convert_seurat_findallmarkers_to_wide(data):
-    """Convert from Seurat FindAllMarkers() format to SCP DE wide format
+def canonicalize_name_and_order(data, header_refmap):
+    """Ensure dataframe has column names and ordering as expected by DE UI
     """
 
-    # Update columns to use SCP's expected names and order
-    data = data.rename(columns={"cluster": "group", "gene": "genes"})
-    data = data.astype({"group": "string"})
-    data = data.assign(comparison_group="rest")
-    expected_header_order = [
-        "genes", "group", "comparison_group", "avg_log2FC", "p_val_adj", "p_val", "pct.1", "pct.2"
-    ]
-    data = data[expected_header_order]
+    is_one_vs_rest_only = header_refmap["comparison_group"] == "None"
+    if is_one_vs_rest_only:
+        data = data.assign(comparison_group="rest")
 
-    wide_df = convert_long_to_wide(data)
-    return wide_df
+    # Update column names in DF to expected header names
+    rename_map = {}
+    for ref_header in header_refmap:
+        if ref_header in ["size", "significance"]:
+            continue
+        actual_header = header_refmap[ref_header]
+        rename_map[actual_header] = ref_header
+    data = data.rename(columns=rename_map)
+
+    data = data.astype({"group": "string"})
+
+    # Update headers to expected order
+    unsorted_headers = list(data.columns)
+    size = header_refmap["size"]
+    significance = header_refmap["significance"]
+    sorted_headers = sort_headers(unsorted_headers, size, significance)
+    canonical_long_df = data[sorted_headers]
+
+    return canonical_long_df
 
 
 def convert_long_to_wide(data):
-    """Convert from long format to SCP DE wide format
+    """Convert from canonical DE long format to SCP DE wide format
 
     (Long format is typical uploaded, but this module internally uses wide.)
     """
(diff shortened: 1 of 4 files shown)


bistline (approved):
Looks good!

ingest/author_de.py
```diff
@@ -468,42 +477,55 @@ def generate_result_files(self, col, genes, rest, groups, clean_val, metrics):
...
         comparisons = comparisons_dict.keys()
         rows_by_comparison = dict.fromkeys(comparisons, [])
+
         for comparison_metrics in grouped_comparison_metrics:
             metric_values = []
             for comparison_metric in comparison_metrics:
 
                 # E.g. "B cells--CSN1S1 macrophages"
-                comparison = check_group(comparisons_dict, comparison_metric)
+                comparison = '--'.join(comparison_metric.split('--')[0:2])
```
jlchang, line 486: Thanks for detecting and fixing the `check_group` bug!

ingest/author_de.py
```diff
@@ -47,32 +35,45 @@ def sort_comparison(groups):
...
+    """Convert from canonical DE long format to SCP DE wide format
 
     (Long format is typical uploaded, but this module internally uses wide.)
     """
     metrics = list(data.columns[3:])
     data["combined"] = data["group"] + "--" + data["comparison_group"]
     frames = []
     # E.g.
-    # gene  group   comparison_group    log2foldchange  pvals_adj   qvals   mean    cat dog
+    # genes  group   comparison_group    log2foldchange  pvals_adj   qvals   mean    cat dog
```
jlchang, line 76: Should the column headers either all be singular or all be plural?

singular: gene group comparison_group log2foldchange pval_adj qval mean cat dog

plural: genes groups comparison_groups log2foldchanges pvals_adj qvals means cats dogs

Most internet sources seem to advocate for singular unless the column in question allows multiple values (ie an array). I found enough of the reasons listed in [this StackOverflow thread](https://stackoverflow.com/questions/338156/table-naming-dilemma-singular-vs-plural-names) to be intuitive that my personal preference would be for singular but there are also reasonable arguments for plural. I suggest choosing one form and being consistent.

jlchang (approved):
One non-blocking suggestion for the column headers.

ingest/author_de.py
eweitz: Astute point!  That SO thread is insightful.  Per chat, it seems noteworthy that Seurat uses singular, whereas Scanpy uses plural.  I've gone ahead and updated "genes" to "gene" in the canonical output names, and column header value.  

This will make the transformed output in that column header more consistent, and also make our corresponding internal CLI arguments more consistent.

PR #324, opened 2023-11-14
Fix breaking changes related to upstream OLS4 update (SCP-5359)

eweitz:
This updates some OLS API processing and related tests, to fix an upstream breaking change due to a major update in `/ols/api` to OLS4.  That break in turn causes metadata validation to fail, which ultimately breaks visualization initialization in the study creation journey.

More context is [in GitHub](https://github.com/EBISPOT/ols4/issues/538#issuecomment-1755429407) and [in Slack](https://broadinstitute.slack.com/archives/CBEHTH601/p1699982927160689).

This satisfies SCP-5359.


bistline (approved):

jlchang (approved):

jlchang (changes requested):
To fix the test failure, `scp-ingest-pipeline/tests/data/bq_test.json` needs to be updated.
From scp-ingest-pipeline/ingest/validation directory, run:
`python3 metadata_validation.py --bq-json ../../tests/data/annotation/metadata/convention/valid_array_v2.1.2.txt`

This will generate `addedfeed000000000000000.json` which should replace the existing `scp-ingest-pipeline/tests/data/bq_test.json` file

eweitz:
Thanks!  Yes, I fixed that late yesterday (be8533c) and just pushed the change.

There's also at least 1 other test failure, and I'm working on that now.  I've marked the PR as draft to reflect that and will re-request review when the code passes all tests.

bistline (approved):

PR #326, opened 2023-11-15
Restore curl in Docker image; removed in upstream dependency (SCP-5359)

eweitz:
This fixes failing Google Cloud Build runs, which broke due to our Docker image's base dependency `marketplace.gcr.io/google/ubuntu2004:latest` having removing `curl`.

This relates to SCP-5359.


jlchang (approved):
Curse you, `latest`! Welcome back `curl`

bistline (approved):

PR #332, opened 2023-12-04
Increasing maximum number of MongoDB retries, jitter for backoff (SCP-5429)

bistline:
#### BACKGROUND & CHANGES
This update doubles the maximum number of retries for MongoDB inserts from 5 to 10, and increases the maximum amount of jitter on exponential backoffs by a factor of 2.5.  This is follow-on work for #331 as the previous fixes were insufficient for ingesting very large matrix files w/ extremely dense data (small slices of the parent matrix were able to ingest, but the entire merged file still failed reproducibly).  The combined updates proved successful for ingesting the matrix at least once (second try in underway).  

The logic is that when a MongoDB connection is reset, the issue is less about the time it takes to reconnect as opposed to the total number of attempts.  We do not differentiate between `AutoReconnect` and `BulkWriteError` when retrying, and both errors tend to travel together.  As such, `serverSelectionTimeoutMS` and batch size decreases only addressed part of the problem.  Additionally, waiting longer between retries allows for more time to stabilize the connection.

Also, this changes the reference for the constant on retries to `MongoConnection.MAX_AUTO_RECONNECT_ATTEMPTS` for both consistency and testing.

#### MANUAL TESTING
Since manual testing involves parsing a ~110 GB dense matrix, this is not advised (see timing information below).  

```
Your Single Cell Portal parse job has completed with the following results:

Total parse time: 21 Hours, 48 Minutes and 37 Seconds
Gene-level entries created: 18677
```

### DIFF
--- a/ingest/mongo_connection.py
+++ b/ingest/mongo_connection.py
@@ -56,8 +56,8 @@ def graceful_auto_reconnect(mongo_op_func):
 
     def retry(attempt_num):
         if attempt_num < MongoConnection.MAX_AUTO_RECONNECT_ATTEMPTS - 1:
-            exp_backoff = pow(2, attempt_num)
-            max_jitter = math.ceil(exp_backoff * 0.5)
+            exp_backoff = pow(2, attempt_num + 1)
+            max_jitter = math.ceil(exp_backoff * 0.2)
             final_wait_time = exp_backoff + random.randint(
                 0, max_jitter
             )  # exponential back off


jlchang (approved):
The new changes seem logical, given the large dataset issues. I'm glad that these changes seem to suffice.

Re: large datasets and Mongo disk space
Is real-time (just hourly-ish) monitoring of available disk or notification of large dataset ingest currently done Rails-side?
Do we think we need any monitoring/alerting from the ingest side to suggest checking available disk based on notably large dataset ingest?

bistline:
> The new changes seem logical, given the large dataset issues. I'm glad that these changes seem to suffice.
> 
> Re: large datasets and Mongo disk space Is real-time (just hourly-ish) monitoring of available disk or notification of large dataset ingest currently done Rails-side? Do we think we need any monitoring/alerting from the ingest side to suggest checking available disk based on notably large dataset ingest?

Great question - we do have real-time monitoring available in the Observability tab for any of our GCE instances.

Production: https://console.cloud.google.com/compute/instances/observability?project=broad-singlecellportal&tab=observability&pageState=(%22nav%22:(%22section%22:%22disk%22))
Staging: https://console.cloud.google.com/compute/instancesDetail/zones/us-central1-a/instances/singlecell-mongo-02?project=broad-singlecellportal-staging&tab=monitoring&pageState=(%22observabilityTab%22:(%22mainContent%22:%22metrics%22,%22section%22:%22diskCapacity%22),%22duration%22:(%22groupValue%22:%22PT1H%22,%22customValue%22:null))

Note: the Ops agent is still installing in staging so the view/data may not be available yet but it should be soon.

ingest/mongo_connection.py
```diff
@@ -52,13 +52,12 @@ def graceful_auto_reconnect(mongo_op_func):
     import random
     import math
 
-    MAX_ATTEMPTS = 5
     # Adopted from https://stackoverflow.com/questions/46939285
 
     def retry(attempt_num):
-        if attempt_num < MAX_ATTEMPTS - 1:
+        if attempt_num < MongoConnection.MAX_AUTO_RECONNECT_ATTEMPTS - 1:
             exp_backoff = pow(2, attempt_num)
```
eweitz, line 59: AIUI, an issue this PR addresses is that the backoff period was too brief, not that retries were too synchronized.  

So given no signs of [thundering herds](https://en.wikipedia.org/wiki/Thundering_herd_problem) per se, I suspect we could more simply -- and maybe more effectively -- mitigate failed large ingests by increasing the underlying backoff, while leaving jitter unchanged.  The change's wider jitter does increase backoff, which indeed ought to help, just more subtly and indirectly.

Below is a trivial way we could implement that.  The cost to test it would be non-trivial, though, so the refinement suggested below doesn't strike me as worthwhile _unless_ we see future similar failures of large ingests.

```suggestion
            exp_backoff = pow(2, attempt_num + 1)
```

eweitz (approved):
The robustness improvements look good!  I suggest a future possible refinement.

> when a MongoDB connection is reset, there is (apparently) no "timeout" that is invoked - the connection is dropped and the `AutoReconnect` exception triggers immediately

Thanks, this is useful to know.  Looking closer into the meaning of [our new custom setting for `serverSelectionTimeoutMS`](https://github.com/broadinstitute/scp-ingest-pipeline/pull/331/files#diff-9c2780f0f58688e63d612ce41491ff85c70167437bce7aa42305e3230dd45589R37), I see [MongoDB docs](https://www.mongodb.com/docs/manual/reference/connection-string/) note it "Specifies how long (in milliseconds) to block for server selection before throwing an exception. Default: 30,000 milliseconds."

So it seems this updated backoff will increase _time between the initiation_ of reconnection attempts, whereas that timeout increased the _max duration_ of a reconnection attempt.  If MongoDB resources are saturated, I can see how increasing wait times _between_ reconnection attempts (as this PR does) would help robustness in addition to a increased wait _within_ each attempt.

bistline:
> Specifies how long (in milliseconds) to block for server selection before throwing an exception. Default: 30,000 milliseconds.

Thanks for the insightful comment.  That is indeed correct, and I'm realizing now that my comment is slightly incorrect/misleading.  

What I've found through testing is that the `serverSelectionTimeoutMS` was only a small part of the overall issue.  When the connection is dropped, the ingest process is finding the MongoDB server again, but that connection keeps getting dropped/reset.  Why is still somewhat of a mystery to me, but it does correlate with large inserts.  Reducing the batch size helped reduce these, but not completely.  Same with the timeout - it made the individual connection retries more resilient, but instances where it couldn't find the server in enough time are much rarer than it finding the server only to have the connection dropped again.  Also, we don't differential between the `AutoReconnect` exception and the `BulkWriteError` in the same retry logic.  These two very often travel together.  By doubling the number of retries, this makes the entire process much more resilient.  The jitter increase was less about "thundering herd" and more "just wait longer".  I think your suggestion of just adding `1` makes more sense than increasing the jitter coefficient, so I'll incorporate that change.

PR #337, opened 2024-03-14
Index ATAC-seq fragment files, enabling visualization in igv.js (SCP-5529)

eweitz:
This prototypes a module to enable displaying indexed fragment files from the Pipelines team in IGV in a web browser.

### Test
1.  Download fragments file from [GCP bucket](https://console.cloud.google.com/storage/browser/fc-bcc55e6c-bec3-4b2e-9fb2-5e1526ddfcd2/reference_data_dev/nemo_atac/multiome_dsp_pipelines?pageState=(%22StorageObjectListTable%22:(%22f%22:%22%255B%255D%22))&prefix=&forceOnObjectsSortingFiltering=false)
2. Run `which bgzip` and `which tabix`, to check if you have those tools installed
3. If above output reports tools are already installed -- it'd print something like `/opt/homebrew/bin/tabix`. If step 2 commands return nothing -- then run `brew install htslib`
4. Re-run step 2 to confirm tools installed
5. `cd ingest`
6. `cp -p <fragments file from GCP bucket> .`
7. `python atac.py`
8. Run `ls -latr *bed*`
9. Confirm output includes .bed.gz and .bed.tbi files

More context is available in this week's [SCP demo recording](https://drive.google.com/file/d/1FNPSXTIQ1wWGPqFu4fA0B0TXgpPv68A3/view?t=23s), including how to visualize these in a standalone IGV web app.  This satisfies SCP-5529.

### DIFF
--- a/ingest/atac.py
+++ b/ingest/atac.py
@@ -2,6 +2,7 @@
 
 PREREQUISITE:
 - Install htslib: `brew install htslib` on macOS (TODO: Create Docker file)
+- Install bedtools: `brew install bedtools` on macOS (TODO: Create Docker file)
 
 """
 import os
@@ -20,12 +21,8 @@ def index_fragments(tsv_path):
     os.rename(tmp_tsv_path, bed_path)
 
     print("Sort BED file by genomic position")
-    # The "-k1.4V" argument ensures `sort` uses column "1" and breaks ties
-    # by ordering character "4" using version-sort, i.e. "V", which does a
-    # natural sort of (version) numbers within text.  This ensures e.g. "12"
-    # appears after "2".
     sorted_bed_path = bed_path.replace(".bed", ".possorted.bed")
-    sort_command = (f"sort -k1.4V {bed_path}").split(" ")
+    sort_command = (f"bedtools sort -i {bed_path}").split(" ")
     sorted_bed_file = open(sorted_bed_path, 'w')
     subprocess.call(sort_command, stdout=sorted_bed_file)
 
@@ -38,6 +35,8 @@ def index_fragments(tsv_path):
     # tabix creates an index for the tab-delimited files, used for getting small chunks
     tabix_command = (f"tabix -s 1 -b 2 -e 3 {sorted_bed_path}.gz").split(" ")
     subprocess.call(tabix_command)
+
+    print("Successfully indexed fragments file for visualization in igv.js")
 
 index_fragments("Library-8-20230710_atac.fragments.tsv")
 
--- a/ingest/atac.py
+++ b/ingest/atac.py
@@ -11,12 +11,13 @@ import subprocess
 def index_fragments(tsv_path):
     """Bgzip and TBI-index a raw ATAC-seq fragments file
     """
-    print("Index fragments file")
+    print("Index fragments file from DSP Pipelines team")
 
     print("Change suffix from .tsv to .bed, as expected by igv.js")
-    shutil.copy2(tsv_path, f"{tsv_path}.tmp") # Preserve original file
-    bed_path = tsv_path.replace(".tsv.tmp", ".bed")
-    os.rename(tsv_path, bed_path)
+    tmp_tsv_path = f"{tsv_path}.tmp"
+    shutil.copy2(tsv_path, tmp_tsv_path) # Preserve original file
+    bed_path = tmp_tsv_path.replace(".tsv.tmp", ".bed")
+    os.rename(tmp_tsv_path, bed_path)
 
     print("Sort BED file by genomic position")
     # The "-k1.4V" argument ensures `sort` uses column "1" and breaks ties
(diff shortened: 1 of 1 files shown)


jlchang (approved):
Functional and code review - looks good!

ingest/atac.py
```diff
@@ -0,0 +1,43 @@
...
+    print("Compress sorted BED file with blocked gzip")
+     # bgzip enables requesting small indexed chunks of a gzip'd file
+    bgzip_command = (f"bgzip -kf {sorted_bed_path}").split(" ")
+    subprocess.call(bgzip_command)
+
+    print("Make a TBI index of the bgzipped BED file")
+    # tabix creates an index for the tab-delimited files, used for getting small chunks
+    tabix_command = (f"tabix -s 1 -b 2 -e 3 {sorted_bed_path}.gz").split(" ")
+    subprocess.call(tabix_command)
+
```
bahill, line 41: The tiniest of requests for the future - log some sort of "success! 🎉 "

ingest/atac.py
```diff
@@ -0,0 +1,43 @@
...
+    bed_path = tmp_tsv_path.replace(".tsv.tmp", ".bed")
+    os.rename(tmp_tsv_path, bed_path)
+
+    print("Sort BED file by genomic position")
+    # The "-k1.4V" argument ensures `sort` uses column "1" and breaks ties
+    # by ordering character "4" using version-sort, i.e. "V", which does a
+    # natural sort of (version) numbers within text.  This ensures e.g. "12"
+    # appears after "2".
+    sorted_bed_path = bed_path.replace(".bed", ".possorted.bed")
+    sort_command = (f"sort -k1.4V {bed_path}").split(" ")
```
bahill, line 28: Is there any advantage to using something like [bedtools](https://daler.github.io/pybedtools/autodocs/pybedtools.bedtool.BedTool.sort.html) for this?

bahill (approved):
even if you want to use bedtools I don't that blocks this PR

ingest/atac.py
eweitz: Thanks for the suggestion!  Yes, `bedtools` has a speed advantage.  And even though it has complexity disadvantage, runtime speed seems more useful than simplicity in this case.  So I've gone ahead and used `bedtools` here (4675c2c).

More specifically, compared to `sort -k1.4V`, `bedtools sort` is:

  * **Faster to run**, by > 60% (14 s) in my local test -- details below.
  * **Harder to debug**.  `bedtools sort` uses lexicographic order, so e.g. chr12 comes before chr2, which is less human-friendly and might confuse someone looking directly in the sorted BED file compared to the more natural order given by `sort -k1.4V`.  But the output is still valid input for `bgzip`, which only requires features to be listed contiguously by chromosome (or scaffold).
  * **An extra dependency to manage**.  Updates, deprecations, and installation are likely to be more expensive with `bedtools`, but probably not by much.

**Speed comparison**
For the 313 MB ATAC-seq fragments file from DSP Pipelines:

```
# bedtools sort
wmce2-f16:atac eweitz$ time bedtools sort -i Library-8-20230710_atac.fragments.bed > Library-8-20230710_atac.fragments.bedtools_sorted.bed

real	0m8.197s
user	0m6.893s
sys	0m0.983s

# DSP Pipelines sort
wmce2-f16:atac eweitz$ time sort -k1,1V -k2,2n Library-8-20230710_atac.fragments.bed > Library-8-20230710_atac.fragments.sort_k11V_k22n.bed

real	0m10.474s
user	0m8.814s
sys	0m1.171s

# Original sort in this PR
wmce2-f16:atac eweitz$ time sort -k1.4V Library-8-20230710_atac.fragments.bed > Library-8-20230710_atac.fragments.sort_k14V.bed

real	0m22.207s
user	0m20.485s
sys	0m1.284s
```

The "DSP Pipelines sort" is from [warp/tasks/skylab/H5adUtils.wdl](https://github.com/broadinstitute/warp/blob/47fb3c33e1dcc9c00e65cd9084ac29857794c2ea/tasks/skylab/H5adUtils.wdl#L271C5-L271C23).  Per data above, `bedtools sort` is > 20% faster than `sort -k1,1V -k2,2n`.  Output seems similar in debuggability; the only disadvantage I see relative to the Pipelines approach is that bedtools would be an extra dependency.  (This might interest you, @ekiernan.)

ingest/atac.py
eweitz: Done, see 4675c2c.  Logging of job status will be handled by Rails once this code is integrated into Ingest Pipeline, but I agree it's good and easy to have a simple success message for now.

For context, you can find that Rails logging [in ingest_job.rb](https://github.com/broadinstitute/single_cell_portal_core/blob/4833ae344ed3ce8c6f4181b31523b131c6e38efb/app/models/ingest_job.rb#L833).  Related analytics reports are mostly in the "[Ingest - Single Cell Portal](https://mixpanel.com/project/2120588/view/19059/app/boards#id=734187)" board in Mixpanel.

PR #338, opened 2024-03-20
Update alexandria metadata convention for BRAIN RF1 (SCP-5549)

jlchang:
Update the Alexandra metadata convention to have boolean metadatum for `has_electrophysiology` and `has_morphology` for BRAIN RF1 work.

Also updating `development_stage` for [Zendesk 318207](https://broadinstitute.zendesk.com/agent/tickets/318207) to include Mouse Developmental Stages ontology in listed ontologies (in addition to Human Developmental Stages ontology).

All above are optional metadata. Translation from `dandi_url` to `has_electrophysiology`=TRUE and `bil_url` to `has_morphology`=TRUE will occur in future ingest code (see SCP-5566) - automated test for the metadata change in this PR will be implemented after the ingest pipeline code is updated. In the meantime, deployment of this metadata convention should not break anything if deployed to production.

This schema change fulfills SCP-5549.


eweitz (approved):

kilonzi (approved):
Good stuff. Thanks

PR #341, opened 2024-04-03
Enable friendly gene search for CELLxGENE AnnData (SCP-5571)

eweitz:
This enables human-friendly gene search for AnnData studies that index by gene ID, not gene name, as in CELLxGENE.

Previously, such studies would be ingested without error, but yield hard-to-use gene search.  Genes could only be searched by Ensembl gene ID, e.g. ENSMUSG00000000001, but not the gene name as is standard in SCP search, e.g. Gnai3.

Here's how the old and new handling look.

https://github.com/broadinstitute/scp-ingest-pipeline/assets/1334561/1d20182f-648c-4322-b3dd-adb284816438

[SCP394 on staging](https://singlecell-staging.broadinstitute.org/single_cell/study/SCP394/cellxgene-anndata-reproducing-scp2557-on-staging) was ingested with this update in place.

This satisfies SCP-5571.


bistline (approved):

bistline:
Also confirmed the new `test_gene_id_indexed_generate_processed_matrix` works locally following the instructions provided.

ingest/anndata_.py
```diff
@@ -155,7 +155,14 @@ def generate_processed_matrix(adata):
         h5ad_frag.features.processed.tsv
         Gzip files for faster delocalization
         """
-        pd.DataFrame(adata.var.index).to_csv(
+        if adata.var.index.name == 'gene_ids':
+            # CELLxGENE indexes by Ensembl gene ID, not gene name (i.e. symbol).
+            # Gene name is encoded in feature_name, which is needed for gene search.
+            feature_frame = adata.var.feature_name
```
jlchang, line 161: Ingest's [mtx processing will extract the Ensembl ID](https://github.com/broadinstitute/scp-ingest-pipeline/blob/79356b955276d016ec8d75d17b57071207a53045/ingest/expression_files/mtx.py#L265) and store it with the gene name (so the gene expression data CAN be searchable by both ensembl ID AND gene name) if we generate a two column features file. 

Leave the original "pd.DataFrame(adata.var.index).to_csv" for the else clause and give something like this a try in the if clause:

df[['var_names', 'feature_name']].to_csv(
            "h5ad_frag.features.processed.tsv.gz",
            sep="\t",
            index=False,
            header=False,
            compression="gzip",
        )

jlchang (changes requested):

ingest/anndata_.py
eweitz: That approach didn't work on my end, but (after some trial-and-error) [preserving the index](https://github.com/broadinstitute/scp-ingest-pipeline/pull/341/commits/17f23e9abb3410386f8a736d2b0b12098920adaf) did.

Per chat, search by gene ID hasn't worked in a long time, so this won't add user value right now.  But, when that's restored, this handling might be needed for alternative search if a study uses a bespoke gene annotation.

I've updated the automated test, and confirmed by manually uploading again to the study on staging.

jlchang (approved):
Thanks for accepting my change request, @eweitz. I've manually confirmed that a two column features file is generated when using `gcr.io/broad-singlecellportal-staging/scp-ingest-ew-cellxgene-anndata:12f80e4 `. 

I'm glad AnnData going forward will store gene_ids in addition to gene_names, eschewing an accumulation of technical debt relative to mtx ingest behavior.

(Apologies that I failed to mention that anndata.var_names was an alias for anndata.var.index - I hope the trial and error didn't take too long)

eweitz:
> anndata.var_names was an alias for anndata.var.index

Makes sense!  I figured originally suggested approach

`df[['var_names', 'feature_name']].to_csv(...)` 

also implied constructing a DataFrame `df`.  But while `df = pd.DataFrame(adata.var.index, adata.var.feature_name)` produced a DataFrame, calling `df.to_csv(<suggested parameters>)` didn't output the expected features file .  Perhaps it would with a different `df` initialization, or different `to_csv` parameters.

PR #353, opened 2024-07-24
Cell Annotation Service: proof of concept

eweitz:
This is a "Hello World" for Cellarium CAS in SCP.  [Cell Annotation Service](https://www.cellarium.ai/tool/cellarium-cell-annotation-service-cas/) is a powerful, convenient way to assign cell types.

The simple kernel here demonstrates basic CAS-SCP plumbing.  The CAS [quickstart_tutorial.ipyb](https://github.com/cellarium-ai/cellarium-cas/blob/main/notebooks/quickstart_tutorial.ipynb), [repo](https://github.com/cellarium-ai/cellarium-cas), and [docs](https://cellarium-cas.readthedocs.io/) are useful resources.

### Test
The manual test below is especially optional, and mainly for reference.  Until we set up Google Secrets Manager (GSM) here, you'll need to use your own CAS API token.  It's easy to get, but turnaround might not be instant.

1. Get a CAS API token.  Visit https://cellarium.ai/tool/cellarium-cell-annotation-service-cas, scroll to bottom, sign up.  
2. Create a file `.cas-api-token` in `scp-ingest-pipeline/ingest/`, paste token, save
3.  Download https://cf.10xgenomics.com/samples/cell-arc/2.0.0/pbmc_unsorted_3k/pbmc_unsorted_3k_filtered_feature_bc_matrix.h5 to `scp-ingest-pipeline/ingest/`
4. `pip install -r requirements.txt`
5. `cd ingest`
6. `python cas.py`
7. Confirm you see output like:

```
* [09:40:49.644] Connecting to the Cellarium Cloud backend with session 94700f52-f5fb-4c25-87a1-18fb26e0d9bd...
* [09:40:50.011] User is eweitz
* [09:40:51.552] Authenticated in Cellarium Cloud v. 1.4.4
* [09:40:51.552] Allowed model list in Cellarium CAS:
  - cellarium-pca-default-v1 (default)
    Description: 64 dimensional PCA Model trained with library size normalization, log1p, zscore with a highly variable gene selection.
    Schema: refdata-gex-GRCh38-2020-A
    Embedding dimension: 64

* [09:40:52.663] Cellarium CAS (Model ID: cellarium-pca-default-v1)
* [09:40:52.664] Total number of input cells: 3009
* [09:40:55.960] The input data matrix has 81156 extra features compared to 'refdata-gex-GRCh38-2020-A' CAS schema (36601). Extra input features will be dropped.
[...]

PR #354, opened 2024-08-01
Enable merging cell types from CAS into AnnData via CLI (SCP-5732)

eweitz:
This supports adding ontology-aware [CAS](https://www.cellarium.ai/tool/cellarium-cell-annotation-service-cas/) cell type annotations into an AnnData object that can initialize an SCP study.

Here's how the resulting study looks!

https://github.com/user-attachments/assets/d8978879-661e-44a7-8399-23534bd6d91c

[Previously](https://github.com/broadinstitute/scp-ingest-pipeline/pull/353), we could call CAS for an AnnData file and print results to the terminal.  Now, we can integrate CAS results into the AnnData file in a manner that can be usefully searched and interactively visualized in a study.  

The full CAS results JSON is also saved, as are interstitial copies.  The CAS JSON seems useful to store in a bucket.  The interstitial AnnData is mostly a helpful diagnostics tool that we can discard as this matures, if we desire.

### Test
1.  Do steps from [PR #353](https://github.com/broadinstitute/scp-ingest-pipeline/pull/353)
2. Confirm output file exists: `cas_annotated_trimmed_compliant__pbmc_10x_v3_4k.h5ad`
3. Create a study
4. Upload `cas_annotated_trimmed_compliant__pbmc_10x_v3_4k.h5ad` as an AnnData file
5. Confirm annotations like "cell_type_cas_1", "cell_type_cas_2", and "cell_type_cas_3" appear
6. Confirm search and interactive visualizations work as expected

This satisfies SCP-5732.


eweitz:
Good questions!  I'd start with considering the simplest way to engineer CAS merge for AnnData as part of ingest.  That seems like it'd be to merge CAS results into the user's AnnData file, and then completely re-ingest the updated AnnData file.

That would show CAS results in the AnnData file users download.

We could also provide the full CAS JSON for user download, e.g. to let users see confidence scores in a notebook.  If we learn that updating the author's downloadable AnnData file is undesirable, then we could point them to the `merge_cas_results` function in this code for how to get CAS results into the original AnnData file using the CAS JSON.

There's a fair number of additional technical questions we'll need to answer as we integrate CAS into ingest.  I've added it as a [topic](https://docs.google.com/document/d/11eNlwMjwKoUcx4TQflzxw4DuN9q03fNNLphUuAVNIc4/edit#bookmark=id.bq2jscoq0lzk) for SCP engineering strategy.

ingest/cas.py
```diff
@@ -3,24 +3,204 @@
...
+        cl=cl,
+        min_acceptable_score=0.1,  # minimum acceptable score for a call
+        cluster_label_obs_column='cluster_label',  # .obs column containing cluster labels
+        top_k=3,  # how many top calls to make?
+        obs_prefix='cell_type_cas'  # .obs column to write the top-k calls to
+    )
+
+    return adata
+
+def trim_cas_adata(adata):
```
jlchang, line 133: As a study owner, I'd probably be interested in the score.... For SCP, we may want to take a look at scores and ask CAS-interested SCP users when/how they would use thresholding with scores to interpret CAS results.

ingest/cas.py
```diff
@@ -3,24 +3,204 @@
...
+#     with open(cas_response_filepath) as f:
+#         cas_response = json.loads(f.read())
+
+#     tsv_rows = []
+#     for cas_item in cas_response:
+#         cell = cas_item["query_cell_id"]
+#         first_match = cas_item["matches"][0]
+#         annotation_label = cas_item["matches"][0]
+
+def merge_cas_results(adata, cas_ontology_aware_response):
```
jlchang, line 96: For the study owner-requested CAS use case, merging definitely makes sense. Exploring extracting CAS results as a clustering file is another avenue, especially if we find a UX that involves enabling user-requested CAS.

jlchang (approved):

ingest/cas.py
eweitz: That makes sense!  The [CAS notebook](https://github.com/cellarium-ai/cellarium-cas/blob/main/notebooks/quickstart_tutorial.ipynb) has a visualization that enables users to see CAS cell type assignments in a tree plot.  It colors cell type nodes by score, has sliders to set score thresholds, etc.  If we ask users about incorporating scores into CAS-SCP views, that's one option to consider.

ingest/cas.py
eweitz: Agreed, this seems worth more exploration in SCP-5753 and / or [CAS-SCP engineering strategy](https://docs.google.com/document/d/11eNlwMjwKoUcx4TQflzxw4DuN9q03fNNLphUuAVNIc4/edit#heading=h.elzs70nakpsf).

eweitz:
Done!  It's now in https://console.cloud.google.com/storage/browser/fc-2f8ef4c0-b7eb-44b1-96fe-a07f0ea9a982/test_Data/scp-ingest-pipeline/cas.

PR #356, opened 2024-08-06
Replacing vault with GSM for local development (SCP-5716)

bistline:
#### BACKGROUND & CHANGES
This replaces all existing Vault integration with Google Secrets Manager (GSM).  Similar to [this PR](https://github.com/broadinstitute/single_cell_portal_core/pull/2049) in `single_cell_portal_core`, a migration script has been added to copy relevant secrets out of Vault and into GSM.  However, since the aforementioned PR moved some of the required data already, this new migration script is agnostic and will allow the user to copy or roll back a single provided secret, rather than multiple hard-coded ones.  Additionally, all READMEs and local setup scripts for native Python or Docker have been updated to reference this new architecture.

#### MANUAL TESTING
1. If you are not authenticated into gcloud, run `gcloud auth login` and make sure you have your development SCP GCP project loaded:
```
gcloud info --format="value(config.project)" # should return broad-singlecellportal-{broad username}
```
2. Pull branch and the run the migration script to copy `mongo/hostname` over to GSM:
```
scripts/vault-gsm-migration.sh -t <path to your vault token> -s mongo/hostname
```
3. Once completed, run `source scripts/setup-mongo-dev.sh`
4. Run `env` and ensure values for `DATABASE_HOST` and `MONGODB_PASSWORD` have been set
5. (Optional) Start Docker Desktop, then run the two setup scripts to validate you get a running Docker container with the latest version of development:
```
scripts/ingest-local-setup.sh && scripts/docker-compose-setup.sh
```

### DIFF
--- a/scripts/ingest-local-setup.sh
+++ b/scripts/ingest-local-setup.sh
@@ -4,25 +4,14 @@
 #
 # Keep "Dev env vars" synced with `setup-mongo-dev.sh`
 
-VAULT_TOKEN_PATH="$1"
-if [[ -z "$VAULT_TOKEN_PATH" ]]
-then
-  echo "You must provide a path to a GitHub token to proceed, e.g. ~/.github-token"
-  exit 1
-fi
-vault login -method=github token=$(cat $VAULT_TOKEN_PATH)
-if [[ $? -ne 0 ]]
-then
-  echo "Unable to authenticate into Vault"
-  exit 1
-fi
+GOOGLE_PROJECT=$(gcloud info --format="value(config.project)")
 
 # Dev env vars
 BROAD_USER=`whoami`
 MONGODB_USERNAME='single_cell'
 DATABASE_NAME='single_cell_portal_development'
-MONGODB_PASSWORD=`vault read secret/kdux/scp/development/$BROAD_USER/mongo/user | grep password | awk '{ print $2 }' `
-DATABASE_HOST=`vault read secret/kdux/scp/development/$BROAD_USER/mongo/hostname | grep ip | awk '{ $2=substr($2,2,length($2)-2); print $2 }' ` 
+MONGODB_PASSWORD=$(gcloud secrets versions access latest --project=$GOOGLE_PROJECT --secret=mongo-user | jq .password)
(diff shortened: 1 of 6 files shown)


scripts/ingest-local-setup.sh
```diff
@@ -4,25 +4,14 @@
...
+GOOGLE_PROJECT=$(gcloud info --format="value(config.project)")
 
 # Dev env vars
 BROAD_USER=`whoami`
 MONGODB_USERNAME='single_cell'
 DATABASE_NAME='single_cell_portal_development'
-MONGODB_PASSWORD=`vault read secret/kdux/scp/development/$BROAD_USER/mongo/user | grep password | awk '{ print $2 }' `
-DATABASE_HOST=`vault read secret/kdux/scp/development/$BROAD_USER/mongo/hostname | grep ip | awk '{ $2=substr($2,2,length($2)-2); print $2 }' ` 
+MONGODB_PASSWORD=$(gcloud secrets versions access latest --project=$GOOGLE_PROJECT --secret=mongo-user | jq .password)
+DATABASE_HOST=$(gcloud secrets versions access latest --project=$GOOGLE_PROJECT --secret=mongo-hostname| jq -r '.ip[0]')
```
jlchang, line 14: Running the script gives this error:
`jq: error (at <stdin>:3): Cannot index string with number`

jlchang (changes requested):
Not jq-savvy but I think we have a blocking issue with setting `DATABASE_HOST`?

scripts/ingest-local-setup.sh
bistline: Interesting... in looking at what got imported into GSM, your value for `mongo-hostname` is an array as a quoted string, i.e. `"[12.34.56.789]"`, as opposed to an array literal of `["12.34.56.789"]`.  However in Vault, the data is correct, so it's not the source.  Let me do some debugging and see what the issue is.

bistline:
> Not jq-savvy but I think we have a blocking issue with setting `DATABASE_HOST`?

The problem was in fact the source data in Vault, which we've fixed now.  Should be ok.

jlchang (approved):
Thanks for troubleshooting with me on my whacked out Vault value, Jon.
All is good now.

PR #359, opened 2024-08-21
Ingest raw count cells for AnnData files (SCP-5103)

bistline:
#### BACKGROUND & CHANGES
This update adds a new `raw_counts` extraction phase for AnnData files where a list of cell names is inserted into MongoDB as the "raw" cells for this file.  Usually we extract flat files and then use existing ingest classes to process the data.  However, since we only require the cell names for raw counts data, and extracting a full MTX bundle would as such be pointless, this update reads the cell names directly from the AnnData file.  This assumes that there is data in the `adata.raw` slot, and that the cells represented there match those in `adata.obs_names`.  There may be future work required if users are not using that slot, and we may wish to let them specify which slot the raw data is in, or even which index the cell names are.  But for now, the slots are hard-coded.  This is part of work to enable downstream portal actions, such as automated differential expression calculation for AnnData files.

#### MANUAL TESTING
1. Initialize your dev environment as normal
2. Run the example command for AnnData raw counts extraction:
```
python ingest_pipeline.py  --study-id 5d276a50421aa9117c982845 --study-file-id 5dd5ae25421aa910a723a337 ingest_anndata --ingest-anndata --anndata-file ../tests/data/anndata/trimmed_compliant_pbmc3K.h5ad  --extract "['raw_counts']"
```
3. Ensure no error was thrown: 
```
% echo $?
0
```
4. In `log.txt`, look for the following message:
```
2024-08-21T11:42:14-0400 ingest_files INFO:Extracted 1 DataArray for 5dd5ae25421aa910a723a337:h5ad_frag.matrix.raw.mtx.gz Cells
```

### DIFF
--- a/ingest/expression_files/expression_files.py
+++ b/ingest/expression_files/expression_files.py
@@ -101,6 +101,13 @@ class GeneExpression:
         QUERY = {"_id": study_file_id, "study_id": study_id}
 
         study_file_doc = list(client[COLLECTION_NAME].find(QUERY)).pop()
+        # special handling of non-reference AnnData files to always return false
+        # this will allow normal extraction of expression data as raw count cells are already ingested during
+        # the "raw_counts" extract phase
+        if (study_file_doc["file_type"] == "AnnData" and "ann_data_file_info" in study_file_doc.keys() and not
+          study_file_doc["ann_data_file_info"]["reference_file"]):
+            return False
+
         # Name of embedded document that holds 'is_raw_count_files is named expression_file_info.
         # If study files does not have document expression_file_info
         # field, "is_raw_count_files", will not exist.:
--- a/ingest/expression_files/expression_files.py
+++ b/ingest/expression_files/expression_files.py
@@ -104,8 +104,8 @@ class GeneExpression:
         # special handling of non-reference AnnData files to always return false
         # this will allow normal extraction of expression data as raw count cells are already ingested during
         # the "raw_counts" extract phase
-        if (study_file_doc["file_type"] == "AnnData" and "ann_data_file_info" in study_file_doc.keys() and not
-          study_file_doc["ann_data_file_info"]["reference_file"]):
+        if (study_file_doc.get("file_type") == "AnnData" and "ann_data_file_info" in study_file_doc.keys() and not
+          study_file_doc["ann_data_file_info"].get("reference_file")):
             return False
 
         # Name of embedded document that holds 'is_raw_count_files is named expression_file_info.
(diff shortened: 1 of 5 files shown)


jlchang (approved):
Code looks good and manual tests behave as described.

ingest/expression_files/expression_files.py
```diff
@@ -101,6 +101,13 @@ def is_raw_count_file(study_id, study_file_id, client):
         QUERY = {"_id": study_file_id, "study_id": study_id}
 
         study_file_doc = list(client[COLLECTION_NAME].find(QUERY)).pop()
+        # special handling of non-reference AnnData files to always return false
+        # this will allow normal extraction of expression data as raw count cells are already ingested during
+        # the "raw_counts" extract phase
+        if (study_file_doc.get("file_type") == "AnnData" and "ann_data_file_info" in study_file_doc.keys() and not
+          study_file_doc["ann_data_file_info"].get("reference_file")):
```
eweitz, line 108: Easier reasoning:

```suggestion
        if (
          study_file_doc.get("file_type") == "AnnData" and 
          "ann_data_file_info" in study_file_doc.keys() and 
          not study_file_doc["ann_data_file_info"].get("reference_file")
        ):
```

ingest/ingest_pipeline.py
```diff
@@ -537,6 +540,11 @@ def extract_from_anndata(self):
                 "extract"
             ):
                 self.anndata.generate_processed_matrix(self.anndata.adata)
+
+            if self.kwargs.get('extract') and "raw_counts" in self.kwargs.get(
+                "extract"
+            ):
```
eweitz, line 546: Tidier:

```suggestion
            if self.kwargs.get('extract') and "raw_counts" in self.kwargs.get('extract'):
```

or, less ideally but smoother than current IMO:

```
if (
    self.kwargs.get('extract') and 
    "raw_counts" in self.kwargs.get('extract')
):
```

Alas, [Python lacks elegant optional chaining](https://stackoverflow.com/questions/64285182/optional-chaining-for-python-objects-foo-bar-baz).

eweitz (approved):
Code looks good!  Nice follow-on for https://github.com/broadinstitute/single_cell_portal_core/pull/2113.

I suggest some trivial readability refinements, no blockers.

ingest/ingest_pipeline.py
bistline: Agreed - most of this was me trying to match syntax of what we had before.

PR #362, opened 2024-08-26
Extending differential expression analysis to AnnData files (SCP-5764)

bistline:
#### BACKGROUND & CHANGES
This update adds `h5ad` as a valid `matrix_file_type` for executing differential expression analyses.  Our current differential expression implementation in Scanpy uses a temporary h5ad file created from source plain text files, so allowing the use of user-uploaded AnnData files is very simple.  Aside from updating the CLI, the only changes are to use the `raw` slot from the matrix and to not transpose the data.

#### MANUAL TESTING
1. Initialize your local dev environment as usual
2. Run the example command from `ingest_pipeline.py` for AnnData DE and confirm you see similar output:
```
python ingest_pipeline.py --study-id addedfeed000000000000000 --study-file-id dec0dedfeed1111111111111 differential_expression --annotation-name louvain --annotation-type group --annotation-scope study --matrix-file-path ../tests/data/anndata/trimmed_compliant_pbmc3K.h5ad --matrix-file-type h5ad --annotation-file ../tests/data/anndata/h5ad_frag.metadata.tsv --cluster-file ../tests/data/anndata/h5ad_frag.cluster.X_umap.tsv --cluster-name umap --study-accession SCPdev --differential-expression
...
STATUS after DE [0]
studyAccession: SCPdev
fileSize: 1
trigger: dev-mode
logger: ingest-pipeline
appId: single-cell-portal
file_path ../tests/data/anndata/h5ad_frag.cluster.X_umap.tsv
study_file_id dec0dedfeed1111111111111
```
3. Confirm you do not get an error, and then open the `umap--louvain--0--study--wilcoxon.tsv` file and confirm the first 3 entries are as follows:
```
	names	scores	logfoldchanges	pvals	pvals_adj	pct_nz_group	pct_nz_reference
0	LDHB	29.36	2.091	1.776e-189	2.435e-185	0.9196	0.5019
1	RPS27	27.56	1.005	3.229e-167	1.217e-163	0.9991	0.9896
2	RPS25	26.45	1.076	3.659e-154	7.168e-151	0.9982	0.9689
...
```


jlchang (approved):
Forgot we'd generally be running at the point of ingest so all the fragment files are still available. 

Non-blocking question: For the (admittedly rare) case where the AnnData file was already ingested an the user clicks the checkbox for "I have raw counts in the adata.raw slot", would we re-extract metadata and cluster and generate a raw matrix mtx rather than pass the AnnData object to Scanpy?

bistline:
> Forgot we'd generally be running at the point of ingest so all the fragment files are still available. Manual test works great.
> 
> Non-blocking question: For the (admittedly rare) case where the AnnData file was already ingested an the user clicks the checkbox for "I have raw counts in the adata.raw slot" (In this scenario, the fragment files are not available and would need to be generated), would we re-extract metadata and cluster and generate a raw matrix mtx rather than pass the AnnData object to Scanpy?

Those fragments should actually still exist - we don't clean them up currently.  They've proven too useful 😉 .  If they don't exists, then the job will fail and we'd get an error message to that effect.  In that event, then yeah - we'd want to extract them again, though our automation is going to attempt to re-ingest everything at that point.  I think if we run into this scenario, we can handle it ad-hoc.

Also, it's worth noting that them selecting "Yes" for raw counts on an existing file will not launch DE.  We still have to do that manually.

eweitz (approved):
Code looks good!

PR #364, opened 2024-09-06
Minify ontologies and keep them fresh, for fast metadata validation (SCP-5719)

eweitz:
This lays a foundation for drastically faster ontology label validation.  Once integrated, that means easier upload for users.

### Background
Adding metadata for cells currently requires authors to specify ontology IDs and formal labels for certain annotations.  This enables faceted search across studies, consistent fields for exploration within studies, and potential machine-learning applications.  

Ensuring those IDs and labels are correct requires validation when data is added.  Failing this validation is common in the author journey, and at 15-45 minutes per attempt it's a frustratingly slow experience.  Instant validation would help!

#### Minify ontologies
Previously, such ontology metadata was validated _after_ upload by querying EBI's Ontology Lookup Service (OLS).  Each unique ID-label pair was queried via a remote OLS API endpoint.  Results were cached per ingest run, but needing to upload data first and then querying a remote API wasn't especially fast.

Now, the ontologies used by OLS are available as small static files.  A new pipeline module fetches the latest versions of each ontology from its original source on GitHub.  The original ontology is then slimmed into a minimal TSV and compressed.  This converts ~224 MB in ontology JSON files into 2 MB TSV.GZs containing only the needed data: one row for each class, listing its ID, label, and synonyms.  

The smaller reference data will be quickly loadable into the client.  That unlocks doing this validation _before_ upload, and using in-memory data rather than remote queries.  The validation will speed up immensely -- 15-45 _minutes_ before, 10-15 _seconds_ after.

#### Keep them fresh
The minified ontologies are automatically updated.  When a PR is opened, the new "Minify ontologies" GitHub Action workflow runs `minify_ontologies.py` to fetch the newest versions of each ontology.  Note the new GitHub check at bottom of this PR.  The GHA workflow compares the newest version to the version in the repo, and [commits](https://github.com/broadinstitute/scp-ingest-pipeline/pull/364/commits/978d6f0d95d79a44eb641dd2a21a01a6b3bb3124) the new version if it differs.

### Next steps
These static files (e.g. `mondo.min.tsv.gz`) will be useful when they're integrated into client-side file validation (CSFV):

1.  On page load, fetch these ontologies, load into client memory.  They're a 2 MB transfer in total, so fast to load and small even decompressed and parsed.
2.  Upon selecting AnnData, apply [row-level validation](https://github.com/broadinstitute/single_cell_portal_core/pull/2122) for required metadata validation, using these small static ontologies for instant ID-label mapping

[...]

PR #365, opened 2024-09-12
Refine ontology cache, for fast metadata validation (SCP-5790)

eweitz:
This improves coverage and performance in the ontology cache.  That speeds up author and developer experiences.

### Overview
[Previously](https://github.com/broadinstitute/scp-ingest-pipeline/pull/364), remote ontologies were slimmed and cached.  This unlocked major performance gains in validation, but it was a first draft.  Some ontologies were missing, data files were located among code files, and downstream clients had to frequently reprocess the same ontologies.

Now, the ontology cache is wider, tidier, and faster:
* Uberon and the Cell Ontology are newly cached
* Minified ontologies are cohesively located in their own directory
* Redundant synonyms are omitted

Most substantially: web browsers will only fetch ontology data _once_.  A new `version.txt` file enables downstream client-side file validation / CSFV to robustly cache processed mini-ontologies on the frontend.  So the user's browser only once ever fetches and decompresses a given version of the min.tsv.gz files.  Subsequent page loads fetch data from the client's local [service worker cache](https://web.dev/articles/service-worker-caching-and-http-caching).  Clients query `version.txt` on every upload page load, and fetch new copies only when that occasionally updates.  A forthcoming PR in Core will clarify these logistics.

That speeds up this validation rule by ~2 seconds, helping both users and engineers get stuff done faster.

### Test
An automated test verifies these updates.  To manually test, `gzip -d mondo.min.tsv.gz` and confirm the redundant synonym "disease" ("||disease||") doesn't  appear in the row for MONDO_0000001.

This relates to SCP-5790.


bistline (approved):

bistline:
Build failure looks like it was a timeout - some of these builds have been taking ~[15m](https://app.circleci.com/pipelines/github/broadinstitute/scp-ingest-pipeline/2162/workflows/bb4e0bf6-285b-47b7-b356-19f520cb3e83/jobs/2082) whereas "normal" runs are around [5m](https://app.circleci.com/pipelines/github/broadinstitute/scp-ingest-pipeline/2165/workflows/59f8a592-98eb-4e7d-b509-829825febaa7/jobs/2085).  There's no obvious culprit so I'm assuming it's something upstream in either CircleCI or Github.

jlchang (approved):
Code changes make sense to me!

Relevant to metadata validation, but not specifically to this PR, HCA used to use a controlled vocabulary for "sex" (that we use in our validation). HCA has since switched to an ontology and we probably should too (they also allow the string "unknown"). We'll need to check if we have any studies that use the previously allowed term "mixed" (when the biosample included tissue from more than one donor) and probably convert them to "unknown" <- we should check what HCA did. (added to JIRA backlog as [SCP-5804](https://broadworkbench.atlassian.net/browse/SCP-5804))

We're already getting PATO so there's no additional work needed with this ontology minification work. The twist with "sex" is that HCA is only allowing terms below PATO_0001894 in the PATO hierarchy.

FYI, probably not worth the effort but, cellxgene schema says that only Species terms should be a child of [NCBITaxon:33208](http://purl.obolibrary.org/obo/NCBITaxon_33208), if we didn't want to minify all of NCBITaxon.

eweitz:
Insightful notes!  Those seem like promising future refinements, and useful topics for engineering strategy.

PR #366, opened 2024-09-17
Harmonizing default Mixpanel properties across Rails & ingest (SCP-5798)

bistline:
#### BACKGROUND && CHANGES
This update aims to harmonize some common Mixpanel event properties across `scp-ingest-pipeline` & `single_cell_portal_core`.  These include:

- `action`: Main process being performed by ingest run, like `ingest_cluster` or `differential_expression`
- `referenceAnnDataFile`: T/F indication if this is a "reference" AnnData upload (no data extracted/visualized)
- `trigger`: this already existed in `scp-ingest-pipeline`, but the `bucket` value was never supported, meaning we did not know from ingest if this was an file that used the new "bucket path" feature

This will give us better insight into common Mixpanel events, especially for AnnData files.  We can now see the specific action that failed and pair them with the `errorTypes` array, which contains more detailed error information than a simple exit code.

#### MANUAL TESTING
1. Initialize your environment and go to the `ingest` directory
2. Run the sample command to run an AnnData differential expression job:
```
python ingest_pipeline.py --study-id addedfeed000000000000000 --study-file-id dec0dedfeed1111111111111 differential_expression --annotation-name louvain --annotation-type group --annotation-scope study --matrix-file-path ../tests/data/anndata/trimmed_compliant_pbmc3K.h5ad --matrix-file-type h5ad --annotation-file ../tests/data/anndata/h5ad_frag.metadata.tsv --cluster-file ../tests/data/anndata/h5ad_frag.cluster.X_umap.tsv --cluster-name umap --study-accession SCPdev --differential-expression
```
3. In the output from the process, you should `action` listed now in the properties, which should show up as `differential_expression`:
```
STATUS after DE [0]
studyAccession: SCPdev
fileSize: 1
trigger: dev-mode
logger: ingest-pipeline
appId: single-cell-portal
action: differential_expression <= new property
file_path ../tests/data/anndata/h5ad_frag.cluster.X_umap.tsv
study_file_id dec0dedfeed1111111111111
```


ingest/config.py
```diff
@@ -171,12 +175,15 @@ def study_file(self, study_file_id):
                 self.file_type = self.study_file["file_type"]
                 self.file_size = self.study_file["upload_file_size"]
                 self.file_name = self.study_file["name"]
+                upload_trigger = self.study_file.get("options", {}).get("upload_trigger")
```
eweitz, line 178: Some day we'll get optional chaining in Python!

eweitz (approved):
Code looks good!  Useful observability refinement.

jlchang (approved):
This will make the Mixpanel data so much more useful! Thank you.

PR #377, opened 2024-12-12
Fixing Github ref for minify ontology commits (SCP-5881)

bistline:
#### BACKGROUND & CHANGES
This fixes the step in the `Minify ontologies` action when attempting to commit changes back to the feature branch if there are updates to any ontologies.  While the reference was technically correct, the commits were failing with the same error: `error: src refspec {branch name} does not match any`.  According to the [actions/checkout](https://github.com/actions/checkout?tab=readme-ov-file#Push-a-commit-to-a-PR-using-the-built-in-token) documentation, the correct way to push changes back is to check out the branch with the current ref and then simply call `git push` without specifying a ref.

#### MANUAL TESTING
The [last run of Minify ontologies](https://github.com/broadinstitute/scp-ingest-pipeline/actions/runs/12300527637/job/34328910881) succeeded without needing to add the branch to the "push" block.


eweitz (approved):
Nice diagnosis of the unintuitive cause, and fix!

jlchang (approved):

PR #379, opened 2025-01-23
Extending MongoDB auto-reconnect to all inserts (SCP-2629, SCP-5904)

bistline:
#### BACKGROUND & CHANGES
This update extends the current MongoDB "graceful auto-reconnect" functionality to all `insert_many` calls, not just those specific to gene expression data.  This is in response to some instances of `Connection reset by peer` when ingesting other file types, such as metadata or subsampling of very large files.  Additionally, now all database connections from ingest use the `MongoConnection` module, standardizing how `scp-ingest-pipeline` connects to database instances (as outlined in SCP-2629).

#### MANUAL TESTING
This is (sadly) nearly impossible to test, as it requires a file that is not an expression matrix that will reliably trigger the `Connection reset by peer` error.  Testing with the fragment files from SCP2805 will take multiple days to encounter the error (original job failed after ~5 days of runtime), if it happens at all.

There is an automated test `test_insert_reconnect` in `test_ingest.py` that mirrors the reconnection test in the `GeneExpression` module.  This test validates that when either the `AutoReconnect` or `BulkWriteError` errors occur, the reconnection code fires correctly.  You can run this locally by initializing your ingest environment and then running this command from the `tests` directory:
```
pytest test_ingest.py -k 'test_insert_reconnect' --disable-pytest-warnings
```

You should see output similar to this:
```
=========================================================== test session starts ============================================================
platform darwin -- Python 3.10.14, pytest-7.2.1, pluggy-1.0.0
rootdir: /Users/bistline/Documents/Python/scp-ingest-pipeline
plugins: xdist-1.29.0, dash-2.17.1, forked-1.6.0, cov-2.8.1
collected 26 items / 25 deselected / 1 selected                                                                                            

test_ingest.py .                                                                                                                     [100%]

============================================= 1 passed, 25 deselected, 125 warnings in 21.24s ==============================================

```

Another manual test would be to use the ingest image `gcr.io/broad-singlecellportal-staging/scp-ingest-jb-mongo-retry-all:0f9452d` in your local SCP instance and validate that normal file uploads can write to the database as usual.  Since all DB connections are now standardized and use the same reconnect code, we can at least prove normal operation is uninterrupted.


ingest/mongo_connection.py
```diff
@@ -117,6 +124,7 @@ def discard_inserted_documents(errors, original_documents) -> list[dict]:
     :param original_documents: (list[dict]) list of documents from original transaction that failed
     :returns list[dict]: list of documents with existing entries removed
     """
+    print(original_documents)
```
eweitz, line 127: I assume this is unintended outside development.

```suggestion
```

eweitz (approved):
Code looks good!  Nice consolidation to the more robust approach.

jlchang (approved):
Verified ingest of a known good AnnData file behaves as expected.
Keeping my fingers crossed that this will do the trick for long subsampling jobs!

PR #385, opened 2025-03-12
Update ontology label for metadata validation tests (SCP-5953)

jlchang:
### Background
The official ontology label for HANCESTRO_0005 is now "European ancestry" rather than "European". This change caused several metadata validation tests to fail. Metadata validation inputs and reference validation results have been updated to reflect the new ontology label so that the broken tests will now pass.

### Manual testing
To manually verify that the broken tests have been fixed, go to the `tests` directory of this repo and run:

> pytest


eweitz (approved):
Thanks for the swift update!

PR #386, opened 2025-03-14
Enable `raw_location` input for DE processing

jlchang:


eweitz (approved):
Code looks good!

bistline (approved):
Looks good!

Question: do we want to specify the location with the period/dot (`.`) or just interpolate that in at runtime?  In other words, have the user say `raw` instead of `.raw`?  I'm fine either way, more thinking about collecting that information from the user and having it in a form/data model.

jlchang:
> Question: do we want to specify the location with the period/dot (`.`) or just interpolate that in at runtime? In other words, have the user say `raw` instead of `.raw`? I'm fine either way, more thinking about collecting that information from the user and having it in a form/data model.

Because the keys for layers can be any string, it's possible that someone might decide to call their layer "raw" but they'd be unlikely to call it ".raw". I'm agnostic about how we present it in the upload wizard UI, I just wanted to be able to easily distinguish between the .raw slot and adata.layers['raw'] in the CLI for ingest.

bistline:
> > Question: do we want to specify the location with the period/dot (`.`) or just interpolate that in at runtime? In other words, have the user say `raw` instead of `.raw`? I'm fine either way, more thinking about collecting that information from the user and having it in a form/data model.
> 
> Because the keys for layers can be any string, it's possible that someone might decide to call their layer "raw" but they'd be unlikely to call it ".raw". I'm agnostic about how we present it in the upload wizard UI, I just wanted to be able to easily distinguish between the .raw slot and adata.layers['raw'] in the CLI for ingest.

Ah good call - I sorta glossed over the difference between `adata.raw` and `adata.layers['raw']`.  Let's leave it as is and we can deal with that complexity in the upload wizard.

PR #394, opened 2025-04-15
Making raw_location optional parameter, adding test coverage (SCP-5984)

bistline:
#### BACKGROUND & CHANGES
This addresses a regression with the `raw_location` parameter where all non-AnnData DE jobs fail if this isn't set.  We lacked test coverage for this at the `ingest_pipeline.py` level which is why it wasn't caught.  This update makes the parameter optional, and also adds coverage from the ingest harness for DE invocation.

#### MANUAL TESTING
1. Initialize your dev environment
2. Run `pytest test_ingest.py  --disable-pytest-warnings -k 'test_execute_de'` from the `tests` directory and ensure all DE-based tests pass


ingest/ingest_files.py
```diff
@@ -121,6 +121,7 @@ def __init__(
 
     @staticmethod
     def is_remote_file(file_path):
+        print(f"file_path: {file_path}")
```
eweitz, line 124:
```suggestion
```

eweitz (approved):
Nice diagnosis, collaboration, and fix!  I'm glad we'll have DE restored for classic ingest.

PR #397, opened 2025-06-10
Removing all BigQuery integration (SCP-6020)

bistline:
#### BACKGROUND & CHANGES
This update removes all steps that upload convention metadata to BigQuery.  Metadata will still be validated as normal, but the final step of creating and uploading a BQ-compliant JSON file is omitted as we will no longer be using BigQuery for search purposes.

It is worth noting that while BigQuery is not strictly necessary for search, it served one important purpose in that it replaced any ontology label synonyms with the canonical label before uploading.  This had the effect of reducing the amount of different labels that exists for a given ontology ID.  Without this step, we will instead be using MongoDB to source these labels, which will use whatever synonym the user uploaded.  They will still be valid - and any search will match on the ID, not the label necessarily - but it will have the effect of cluttering the search facet UI.  More work will need to be done in `single_cell_portal_core` to reduce this clutter.

#### MANUAL TESTING
1. Initialize your dev environment and run the sample code to ingest a convention metadata file, confirming it succeeds. (NOTE: you may need to change the study accession if you have SCP123 already):
```
python ingest_pipeline.py --study-id 5d276a50421aa9117c982845 --study-file-id 5dd5ae25421aa910a723a337 ingest_cell_metadata --cell-metadata-file ../tests/data/annotation/metadata/convention/valid_no_array_v2.0.0.txt --study-accession SCP123 --ingest-cell-metadata --validate-convention
...

studyAccession: SCPdev
fileName: 5dd5ae25421aa910a723a337
fileSize: 1
trigger: dev-mode
logger: ingest-pipeline
appId: single-cell-portal
action: ingest_cell_metadata
status: success
functionName: ingest_cell_metadata
perfTime: 0.01
```
2. In your BigQuery console, confirm you have no entries for `SCP123`, or whatever accession you used
```
SELECT COUNT(*) FROM `cell_metadata_development.alexandria_convention`
WHERE study_accession = 'SCP123'

=> 0
```
3. Back in `scp-ingest-pipeline`, run the "old" command with the bq options and confirm you get the following error:
```
python ingest_pipeline.py --study-id 5d276a50421aa9117c982845 --study-file-id 5dd5ae25421aa910a723a337 ingest_cell_metadata --cell-metadata-file ../tests/data/annotation/metadata/convention/valid_no_array_v2.0.0.txt --study-accession SCP123 --ingest-cell-metadata --validate-convention --bq-dataset cell_metadata_development --bq-table alexandria_convention
...

usage: ingest_pipeline.py [-h] --study-file-id STUDY_FILE_ID --study-id STUDY_ID [--profile-memory] [--user-metrics-uuid USER_METRICS_UUID]
                          {ingest_expression,ingest_cell_metadata,ingest_cluster,ingest_subsample,differential_expression,ingest_differential_expression,ingest_anndata,render_expression_arrays,rank_genes}
                          ...
ingest_pipeline.py: error: unrecognized arguments: --bq-dataset cell_metadata_development --bq-table alexandria_convention
```


eweitz (approved):
Nice simplification!

PR #416, opened 2026-02-03
Enable converting AnnData from CELLxGENE to SCP schema (SCP-6096)

eweitz:
A fair amount of LungMap data is in CELLxGENE, which makes it slightly easier to ingest into SCP.  This module transforms an H5AD file from CELLxGENE to use the metadata convention (i.e., schema) used by SCP H5AD files.

The current module works when run manually for the dataset in https://singlecell.broadinstitute.org/single_cell/study/SCP3493.  It could be extended to better integrate with the rest of Ingest Pipeline, so I've put it in `ingest/`

This satisfies https://broadworkbench.atlassian.net/browse/SCP-6096.


ingest/cellxgene_to_scp.py
```diff
@@ -0,0 +1,443 @@
...
+
+def main():
+    """CLI entry point."""
+    import argparse
+    
+    parser = argparse.ArgumentParser(
+        description='Convert CELLxGENE H5AD files to SCP format',
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+        epilog="""
+Examples:
```
bistline, line 418: I'm sure it's ok but just wanted to note the lack of indentation in the formatting here - I'm assuming this is so it prints correctly, yes?

bistline (approved):

ingest/cellxgene_to_scp.py
eweitz: Good question, that's correct!

PR #418, opened 2026-03-03
Enable processing new AnnData, update dependencies (SCP-6104, SCP-6079)

eweitz:
This updates the `anndata` and `scanpy` packages to enable processing AnnData files that researchers generated with newer standard packages.  

It also upgrades old dependencies that hindered maintenance, and removes unused dependencies.


bistline (approved):
Glad to see these packages being updated - thanks for working through all the corners!

PR #422, opened 2026-03-23
handle AnnData files with embedded pandas df (SCP-6118)

bistline:
#### BACKGROUND & CHANGES
This update adds a check to ensure that obtained clustering objects from the `obsm` frame are returned as Numpy `ndarrays` instead of Pandas dataframes.  This is to prevent downstream ingest errors due to mangled flat tsv files.

#### MANUAL TESTING
1. Pull this branch and set up your environment as normal
2. Download the AnnData file from https://www.lungmap.net/dataset/?dataset_id=LMEX0000004388 and confirm you can extract the clustering coordinates to a file with the following command:
```
python ingest_pipeline.py --study-id addedfeed000000000000000 --study-file-id addedfeed000000000000001 --user-metrics-uuid 6a3fc1e6-41a2-4f9b-8ed0-0c75a26217d4 ingest_anndata --ingest-anndata --anndata-file ~/Downloads/LMEX0000004388.h5ad --obsm-keys "['X_umap']" --extract "['cluster']"
```


bistline:
Test failures are due to OLS being down right now (503 for deployment, since we use minified ontologies this doesn't affect ingest but only the one fallback call for non-standard species taxon ids).

eweitz (approved):

eweitz:
I don't recall these particular build failures, before.

bistline:
> I don't recall these particular build failures, before.

https://github.com/EBISPOT/ols4/issues/1209

The whole service is down, so it causes `TestValidateMetadata::test_uses_fallback_validation` to fail since that's the one place where we still (potentially) talk to OLS.  Everywhere else we use the minified ontologies.

